PR #54, opened 2022-06-16
Changes to simplify provisioning EAP 8 builds

TomasHofman:
@spyrkob couple of commits you might consider. (It also includes wildfly-channel upgrade from https://github.com/spyrkob/prospero/pull/53.)


spyrkob:
LGTM, thanks

PR #55, opened 2022-06-17
Implementation of CLI interface implemented in picocli

TomasHofman:
This is a mock CLI interface implemented with picocli, just to show how it could look like.

Related to https://github.com/spyrkob/prospero/issues/33 and other CLI relevant issues.

Screenshot from 2022-06-17 12-32-19


spyrkob:
That's great, thanks. The missing part is calling the actions now right?

TomasHofman:
Yes, they are just mock ups now. I can finish this PR and replace the old commands with these ones, if you are OK with the approach.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
```diff
@@ -0,0 +1,142 @@
...
+                        "or a fully qualified feature pack location like " +
+                        "\"wildfly-core@maven(org.jboss.universe:community-universe):current\"."
+        )
+        Optional<String> fpl;
+
+        @CommandLine.Option(
+                names = "--definition",
+                required = true,
+                // TODO: Fix description.
+                description = "Definition file path. Not sure what this is."
```
spyrkob, line 87: That's a Galleon provisioning XML definition (eg. .galleon/provisioning.xml). It's a way to provision multiple FPs, exclude layers etc

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
```diff
@@ -0,0 +1,142 @@
...
+                    .build();
+
+            Provision provision = actionFactory.install(directory.toAbsolutePath(), mavenSessionManager, console);
+            provision.provision(provisioningDefinition);
+        } catch (ProvisioningException | OperationException e) {
+            console.error("Error while executing installation: " + e.getMessage());
+            return ReturnCodes.PROCESSING_ERROR;
+        }
+
+        return 0;
```
spyrkob, line 136: Should that be `ReturnCodes.SUCCESS`?

prospero-cli/src/test/java/org/wildfly/prospero/cli/CliMainTest.java
```diff
@@ -2,55 +2,41 @@
...
     @Test
-    public void errorIfArgumentHasNoValue() throws Exception {
-        try {
-            new CliMain(actionFactory).handleArgs(new String[]{"install", "--dir"});
-            fail("Should have failed");
-        } catch (ArgumentParsingException e) {
-            assertEquals("Argument value cannot be empty", e.getMessage());
-        }
+    public void errorIfArgumentHasNoValue() {
+        int exitCode = commandLine.execute("install", "--dir");
```
spyrkob, line 23: Can you make the argument names use constants? Just to make it easier if/when we rename them.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
TomasHofman: OK, would "Galleon provisioning XML definition file" be OK description?

Is the definition file a sufficient input to provision a server?

spyrkob: > OK, would "Galleon provisioning XML definition file" be OK description?
Yes, sounds good.

> Is the definition file a sufficient input to provision a server?
It requires the user to provide the channel and repository configuration as well - eg. through `provision-config` parameter.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
```diff
@@ -0,0 +1,142 @@
...
+                // TODO: Check what the correct notations of fpl are.
+                description = "Feature pack location. This can be a well known shortcut like \"eap-8.0\" or \"wildfly\", " +
+                        "or a fully qualified feature pack location like " +
+                        "\"wildfly-core@maven(org.jboss.universe:community-universe):current\"."
+        )
+        Optional<String> fpl;
+
+        @CommandLine.Option(
+                names = "--definition",
+                required = true,
```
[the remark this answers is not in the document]
TomasHofman: In this case it's "exclusive required" - exactly one of the two options must be given.

It's controlled by the `exclusive` attribute on he @ArgGroup annotation. If that would be false, both options would be required:
```
    @CommandLine.ArgGroup(exclusive = true, multiplicity = "1")
    FeaturePackOrDefinition featurePackOrDefinition;
```

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
TomasHofman: OK, makes sense.

prospero-cli/src/test/java/org/wildfly/prospero/cli/CliMainTest.java
TomasHofman: Yes, it was supposed to be in the `OptionNames` class, but then I deleted it to take a proper look at it later. I will add that back.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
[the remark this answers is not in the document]
spyrkob: Ah, sorry missed that. Thanks!

prospero-cli/src/test/java/org/wildfly/prospero/cli/CliMainTest.java
```diff
@@ -2,55 +2,41 @@
...
     @Test
-    public void errorIfArgNameDoesntStartWithDoubleHyphens() throws Exception {
-        try {
-            new CliMain(actionFactory).handleArgs(new String[]{"install", "dir=test"});
-            fail("Should have failed");
-        } catch (ArgumentParsingException e) {
-            assertEquals("Argument [dir=test] not recognized", e.getMessage());
-        }
+    public void errorIfArgNameDoesntStartWithDoubleHyphens() {
+        int exitCode = commandLine.execute("install", "dir=test");
```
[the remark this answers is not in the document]
TomasHofman: I think what you describe is in the later test (errorOnUnknownArgument) and this was supposed to test missing hyphens - fixed.

TomasHofman:
Plus I deleted some YAML files in latest commit that I think are not used anymore.

TomasHofman:
Also I will revisit ordering of options in usage messages.

spyrkob:
Thanks, I'll merge this PR in a moment, the rest can go into new PR

PR #61, opened 2022-06-23
CLI constanst, rename channelsFile references to provisionConfig

TomasHofman:


TomasHofman:
Cc @spyrkob

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/CliConstants.java
```diff
@@ -0,0 +1,34 @@
...
+ * CLI commands and options names
+ */
+public final class CliConstants {
+
+    private CliConstants() {
+    }
+
+    // Commands:
+
+    public static final String MAIN_COMMAND = "prospepo"; // TODO: verify
```
spyrkob, line 13: prospepo/prospero

spyrkob (changes requested):
Looks good, just a typo in one of the Constants

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/CliConstants.java
TomasHofman: Yay, fixed :).

spyrkob:
Thanks!

PR #63, opened 2022-06-23
Move usage strings to a resource bundle, to allow localization

### LINKED ISSUE #33: Add help and syntax description
Add help operation listing available operations and their parameters.

Add --help argument to operations to display the operation's parameter

TomasHofman:
Fixes #33 
Part of #38

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -37,6 +37,7 @@
         <version.org.wildfly.galleon-pack>26.1.0.Final</version.org.wildfly.galleon-pack>
         <version.info.picocli>4.6.3</version.info.picocli>
         <version.system-rules>1.19.0</version.system-rules>
+        <version.org.slf4j>1.7.36</version.org.slf4j>
 
         <version.org.wildfly.checkstyle-config>1.0.8.Final</version.org.wildfly.checkstyle-config>
     </properties>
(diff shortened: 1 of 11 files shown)


TomasHofman:
Cc @spyrkob

TomasHofman:
The install help screen now looks like this: Screenshot from 2022-06-23 14-49-07

TomasHofman:
And the main command help:
Screenshot from 2022-06-24 09-11-19

TomasHofman:
Translations to different languages can be created by adding a language specific property file to the resource bundle, e.g. adding UsageMessages_cs.properties for czech language or UsageMessages_de.properties for german.

For testing purposes, one can use the `-Duser.language=cs` argument to switch the language, as in:

```
java -Duser.language=cs -jar prospero-cli/target/prospero-cli-1.0-SNAPSHOT-shaded.jar
```

Two things remain to figure out for me:
* I don't yet know what determines the default language choice (it's not the LANG env variable).
* Native UTF characters like czech or polish characters with diacritic are not displayed correctly when the property file is UTF-8 encoded. Maybe the file needs to be ASCII encoded. Will test.

spyrkob:
Looks great thanks!

TomasHofman:
One update - native characters work when property files are UTF-8 encoded. My IDE was previously set to use ISO-8859 for property files, which was the problem.
Screenshot from 2022-06-24 09-37-57

TomasHofman:
Update two: the default language choice *is* determined by LC_ALL or LANG env variables. E.g. both `LC_ALL=cs_CZ.UTF-8` and `LANG=cs_CZ.UTF-8` work, when there is given language property file in the resource bundle.

PR #65, opened 2022-06-24
Bulk resolve api

spyrkob:
Incorporates changes to Galleon, galleon-plugins and wildfly-channel using maven artifact bulk resolve operations

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/galleon/ChannelMavenArtifactRepositoryManager.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/galleon/ChannelMavenArtifactRepositoryManager.java
@@ -29,6 +29,7 @@ import org.wildfly.channel.UnresolvedMavenArtifactException;
 
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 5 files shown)


TomasHofman:
Sure, looking.

prospero-common/src/main/java/org/wildfly/prospero/galleon/ChannelMavenArtifactRepositoryManager.java
```diff
@@ -59,70 +57,83 @@ public void resolve(MavenArtifact artifact) throws MavenUniverseException {
...
-                executorService.submit(() -> {
-                    try {
-                        resolve(artifact);
-                        cf.complete(null);
-                    } catch (Exception e) {
-                        cf.completeExceptionally(e);
+                // workaround for when provisioning doesn't add universe artifacts
+                if (artifact.getPath() == null) {
+                    if (isUniverseOrProducerArtifact(artifact.getArtifactId())) {
+                        org.wildfly.channel.MavenArtifact result = channelSession.resolveMavenArtifact(artifact.getGroupId(), artifact.getArtifactId(), artifact.getExtension(),
```
TomasHofman, line 98: Wouldn't this fail with "no stream found", if universe or producer is not part of channel?

prospero-common/src/main/java/org/wildfly/prospero/galleon/ChannelMavenArtifactRepositoryManager.java
```diff
@@ -59,70 +57,83 @@ public void resolve(MavenArtifact artifact) throws MavenUniverseException {
...
         } catch (UnresolvedMavenArtifactException e) {
             throw new MavenUniverseException(e.getLocalizedMessage(), e);
         }
     }
 
-    private DefaultArtifact streamToArtifact(Stream s) {
-        return new DefaultArtifact(s.getGroupId(), s.getArtifactId(), "jar", s.getVersion());
-    }
+    @Override
+    public void resolveAll(Collection<MavenArtifact> artifacts) throws MavenUniverseException {
```
TomasHofman, line 80: I'm asking to be sure - do I assume correctly that this method should be functionally equivalent to the singular resolve() above?

spyrkob: Yes. The outcome of calling resolve on a set of artifacts should be the same as calling resolveAll on that set

TomasHofman:
Apart from the question I posted it looks OK to me :).

TomasHofman (approved):

prospero-common/src/main/java/org/wildfly/prospero/galleon/ChannelMavenArtifactRepositoryManager.java
spyrkob: Yes, it would. In this case we need to handle a case where those artifacts are in the channel, but not the manifest.

The problem scenario is if the server is installed using maven GA it doesn't initially resolve universe or producer, so they're not listed in the manifest. 
If this manifest is then used to re-provision the server (in 'rollback' or 'restore'),  a FPL will be taken from 'provisioning.xml', which might use the universe notation, which would require universe and producer.

TomasHofman: OK, thanks for explanation.

Issue #59, opened 2022-06-22
Input validation - decline to provision into a non-empty directory?


TomasHofman:
If I run `prospero install --dir=eap ...` and the `eap/` directory is non-empty (e.g. contains an old server instance), should prospero report an error, as a safety measure?

TomasHofman:
It currently appears to perform an update, according to `history` command, although the command that was run was `install` in both cases:

```
[thofman@rama prospero]$ java -jar prospero-cli/target/prospero-cli-1.0-SNAPSHOT-shaded.jar history --dir wf-core
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[be5d0121] 2022-06-22T09:08:12Z - update
[524f1f3b] 2022-06-22T09:06:17Z - install
```

spyrkob:
Yes this should be an error.

TomasHofman:
I'm going to open a PR for this.

PR #66, opened 2022-06-27
Decline to install into a non-empty directory

TomasHofman:
Fixes https://github.com/wildfly-extras/prospero/issues/59

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
@@ -126,6 +126,9 @@ public class InstallCommand extends AbstractCommand {
         } catch (ProvisioningException | OperationException e) {
             console.error("Error while executing installation: " + e.getMessage());
             return ReturnCodes.PROCESSING_ERROR;
+        } catch (IllegalArgumentException e) {
+            console.error(e.getMessage());
+            return ReturnCodes.INVALID_ARGUMENTS;
         }
 
         return ReturnCodes.SUCCESS;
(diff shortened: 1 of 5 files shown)


prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
```diff
@@ -103,6 +103,17 @@ public Integer call() {
             return ReturnCodes.INVALID_ARGUMENTS;
         }
 
+        if (directory.toFile().isFile()) {
```
spyrkob, line 106: Can you add those checks to `org.wildfly.prospero.actions.Provision` as well? This should be enforced if Prospero is run via GUI as well

spyrkob (changes requested):

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/InstallCommand.java
TomasHofman: Good point. I would rather move them to the common module, than duplicate. I will need to add some message logger (that can be localized) to the common module for that as well.

spyrkob: Alternatively the Provision can throw an exception the CLI will interpret and show appropriate message

TomasHofman: You can check it out @spyrkob. I used jboss-logging to handle message bundles. We could now add `Messages.i18n_${lang}.properties` files to `resources/org/wildfly/prospero/` to provide language translations.

Any `IllegalArgumentException` that comes from `Provision` class would be treated as wrong argument error.

TomasHofman: We could rework the `Messages` class in `prospero-cli` in similar fashion.

Issue #28, opened 2022-02-25
'-y' flag for update command


spyrkob:
Perform update without asking user for confirmation

TomasHofman:
Going to look at this.

PR #67, opened 2022-06-27
Add -y option to the update command

TomasHofman:
Fixes https://github.com/wildfly-extras/prospero/issues/28.

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/CliConsole.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/CliConsole.java
@@ -146,11 +146,11 @@ public class CliConsole implements Console {
 
     @Override
     public boolean confirmUpdates() {
-        getStdOut().print("Continue with update [y/n]: ");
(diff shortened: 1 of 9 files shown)


spyrkob:
This breaks `UpdateCommandTest` - I think just mock expectations need to be changed

TomasHofman:
Should be fixed now.

prospero-common/src/main/resources/org/wildfly/prospero/Messages.i18n_cs_CZ.properties
```diff
@@ -0,0 +1,2 @@
+dirMustBeDirectory=Daná cesta %s je soubor. Je třeba prázdný adresář.
```
spyrkob, line 1: Wrong PR?

TomasHofman: Yes, that's not supposed to be there. Updated.

spyrkob (approved):

PR #70, opened 2022-06-27
[#64] Prospero should be able to downgrade components

### LINKED ISSUE #64: Prospero should be able to downgrade the component.
If prospero should downgrade some component based on the channel, then:

* currently the component is not upgraded, no warning is printed, channel is not respected
* channel should be respected, component should be downgraded. Warning about this downgrade may be useful

spyrkob:
Fixes #64


TomasHofman (approved):
Looks all right to me.

PR #71, opened 2022-06-27
Move strings to message bundles

TomasHofman:
I tried to move most of the strings that looked important.


spyrkob:
Looks great, thanks!

Issue #72, opened 2022-06-27
Fix slf4j warning when running from jboss-modules


spyrkob:
When using prospero provisioned as standalone FP, SLF4J warning is issued:
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```
In shaded jar this is fixed by adding slf4j-nop library

TomasHofman:
I can take a look if you are not in it already.

spyrkob:
Sure. I suspect it might be because other slf4j jars are in a different module, but hasn't tried it yet

PR #74, opened 2022-06-28
Add slf4j-jboss-logging on the classpath, remove slf4j-nop

TomasHofman:
This resolves the issue with slf4j warning about missing binding.

Fixes https://github.com/wildfly-extras/prospero/issues/72.


spyrkob:
Thanks!

Issue #81, opened 2022-06-29
Should the temporary provisioning repos be deleted?


TomasHofman:
I found quite a few repositories in my `/tmp` folder, like `/tmp/provisioning-repo7585379810969367448/`. Are they needed for anything after a server was installed? Should the tool remove them after provisioning is complete?

spyrkob:
They should be removed. IIRC they should be already marked with `deleteOnExit()`, but I guess we can add `MavenSessionManager#close()` to make sure it's removed

PR #82, opened 2022-06-30
Delete temporary provisioning repo after completing

TomasHofman:
I think `deleteOnExit()` only works if the directory is empty.

Fixes https://github.com/wildfly-extras/prospero/issues/81.


spyrkob:
Thanks @TomasHofman

Issue #42, opened 2022-05-25
Add operation to apply and remove one-off patch from a zip


spyrkob:
The one-off zip contains a maven repository with channel definition and updated jars.

The `prospero apply-patch patch.zip` operation registers the new channel and repository, and performs update.

The `prospero remove-patch <ID>` operation de-registers the new channel and repository, and performs update (effectively removing changes from the patch).

The `prospero list-patch` operation lists installed patches.

Issue #41, opened 2022-05-25
Add management operations to list, add and remove maven repositories and channels


spyrkob:
Add operations to manipulate configured repositories and channels

Example operations:
`prospero repositories list`
`prospero repositories add central https://repo1.maven.org/maven2/`
`prospero repositories remove central`

`prospero channels list`
`prospero repositories add https://localhost/channel.yaml`
`prospero repositories remove https://localhost/channel.yaml`

TomasHofman:
I can take a look at this.

spyrkob:
Just FYI I'm working on #42 now, I'll need to add the API to manipulate the config there, so we'll need to sync up on this

PR #83, opened 2022-06-30
Make it possible to update and persist changes to Config

spyrkob:
Part of #42 and #41

cc @TomasHofman

### DIFF
--- a/prospero-common/src/test/java/org/wildfly/prospero/model/ProvisioningConfigTest.java
+++ b/prospero-common/src/test/java/org/wildfly/prospero/model/ProvisioningConfigTest.java
@@ -1,0 +1,76 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.prospero.model;
+
+import org.junit.Assert;
+import org.junit.Test;
+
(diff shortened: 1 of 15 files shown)


prospero-common/src/test/java/org/wildfly/prospero/model/ProvisioningConfigTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Test(expected = IllegalArgumentException.class)
+    public void addRepositoryThrowsErrorIfSameIdDifferentUrl() throws Exception {
+        provisioningConfig.addRepository(new RepositoryRef("existing", "file:///foo.bar"));
+
+        assertFalse(provisioningConfig.addRepository(new RepositoryRef("existing", "file:///different.url")));
+
+        assertThat(provisioningConfig.getRepositories()).containsExactly(
+                new RepositoryRef("existing", "file:///foo.bar")
+        );
```
TomasHofman, line 51: I think these following assert statements would never be executed, as the exception would be thrown before them.

TomasHofman (changes requested):
Looks good to me, just a small thing in a test case.

prospero-common/src/test/java/org/wildfly/prospero/model/ProvisioningConfigTest.java
spyrkob: Yes that's true, forgot to remove those

prospero-common/src/test/java/org/wildfly/prospero/model/ProvisioningConfigTest.java
```diff
@@ -0,0 +1,70 @@
...
+        assertFalse(provisioningConfig.addRepository(new RepositoryRef("existing", "file:///foo.bar")));
+
+        assertThat(provisioningConfig.getRepositories()).containsExactly(
+                new RepositoryRef("existing", "file:///foo.bar")
+        );
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void addRepositoryThrowsErrorIfSameIdDifferentUrl() throws Exception {
+        provisioningConfig.addRepository(new RepositoryRef("existing", "file:///foo.bar"));
```
TomasHofman, line 45: We could check this returned false, to make sure that IllegalArgumentException was not thrown by this line.

spyrkob: Yeah, I removed `expected` and replaced it with `try/catch` and `Assert.fail`. Clunky, but at least it guarantees to check the right thing

spyrkob:
hmm something funny happened with merging... gonna close this and push last change directly

PR #85, opened 2022-07-01
Apply patch

spyrkob:
Part of #42 - apply-patch operation

Needs some more validation, but the PR is getting big enough, so I'll add it later.

@TomasHofman could you review when you have a moment?


TomasHofman:
Sure - I will take a bit more time on this one, will try to do it later today.

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyPatch.java
```diff
@@ -0,0 +1,129 @@
...
+
+        GalleonUtils.executeGalleon(options -> provMgr.provision(provMgr.getProvisioningConfig(), options),
+                mavenSessionManager.getProvisioningRepo().toAbsolutePath());
+
+        metadata.setChannel(maven.resolvedChannel());
+
+        metadata.writeFiles();
+    }
+
+    private List<Channel> mapToChannels(List<ChannelRef> channelRefs) throws MetadataException {
```
TomasHofman, line 118: Just an idea, we could create a conversion method on the ChannelRef class, e.g. `ChannelRef#toChannel()` (non-static), and then do these conversions via streams, like `channelRefs.stream().map(ChannelRef::toChannel()).collect(...)`. Or maybe have some static utility methods to convert a single instance or a whole collection. It's just that I think this method already exists in a different class. Can be a future task...

TomasHofman (approved):
I think let's merge.

Would the patch repo remain in the .installation/ permanently?

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyPatch.java
spyrkob: Yes, good idea. I want to try and extract the whole section creating ProvisionManager and associated resources to utility at some point

spyrkob:
> Would the patch repo remain in the .installation/ permanently?
Ideally the installed artifacts could be resolved from modules (#65) - then the patch repository can go away.

PR #86, opened 2022-07-01
Adding command for managing maven repositories in an installation

TomasHofman:
Cc @spyrkob, proposal for the repository command.


prospero-common/src/main/java/org/wildfly/prospero/actions/MetadataActions.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    /**
+     * Removes a remote maven repository to an installation.
+     */
+    public void removeRepository(String id) throws MetadataException {
+        InstallationMetadata installationMetadata = new InstallationMetadata(installation);
+        ProvisioningConfig prosperoConfig = installationMetadata.getProsperoConfig();
+        Optional<RepositoryRef> repo = prosperoConfig.getRepositories().stream()
```
spyrkob, line 44: Can you put this as a `removeRepository` method in `ProvisioninConfig`? I think this will be used in other places as well.

btw. I think `ProvisioninConfig` should be renamed to `ProsperoConfig`, but that's separate PR, WDYT?

prospero-common/src/main/java/org/wildfly/prospero/test/MetadataTestUtils.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static InstallationMetadata createInstallationMetadata(List<ChannelRef> channels,
+            List<RemoteRepository> repositories) throws IOException, MetadataException {
+        return createInstallationMetadata(createManifest(null), channels, repositories);
+    }
+
+    public static InstallationMetadata createInstallationMetadata(Channel manifest, List<ChannelRef> channels,
+            List<RemoteRepository> repositories) throws MetadataException, IOException {
+        Path installation = Files.createTempDirectory("installation");
```
spyrkob, line 38: should that be `.installation` -  `InstallationMetadata.METADATA_DIR`

prospero-common/src/main/java/org/wildfly/prospero/test/MetadataTestUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> FileUtils.deleteQuietly(installation.toFile())));
```
spyrkob, line 39: How about using TemporaryFolder junit Rule? It could be a constructor argument and we don't need to worry about cleanup

spyrkob (changes requested):
Looks good, just a couple nitpicks

prospero-common/src/main/java/org/wildfly/prospero/test/MetadataTestUtils.java
TomasHofman: Good idea.

prospero-common/src/main/java/org/wildfly/prospero/test/MetadataTestUtils.java
TomasHofman: Right, I copy pasted that bit and didn't notice.

TomasHofman: Allthough, I'm not sure. The parameter is a prefix string for a temp directory. I think it shouldn't start with ".", that would sort of hide it from the developer, in case the directory was not deleted for any reason (like tests terminated with a signal).

prospero-common/src/main/java/org/wildfly/prospero/actions/MetadataActions.java
TomasHofman: +1 for renaming, wanted to suggest it too. Also thought about renaming the `InstallationMetadata#public org.jboss.galleon.config.ProvisioningConfig getProvisioningConfig()` to `#getGalleonProvisioningConfig()`, to distinguish it more clearly in the API.

spyrkob: I'll make those changes when we merge open PRs to avoid conflicts

spyrkob (approved):

TomasHofman:
I think all changes are ready.

Issue #84, opened 2022-06-30
Move current integration tests to commons module


spyrkob:
Possibly create new integration testsuite as well to test executing from CLI and integrating with WFLY

TomasHofman:
Working on this.

TomasHofman:
What about moving all integration tests into a new module, where they could live in separate packages (for CLI, basic API, ...)? We could reuse config files, utilities, etc.

spyrkob:
Yes, the new testsuite should be a separate module. Initially I think only tests from `prospero-cli/src/test/java/integration/` would live there, but we'll add more

PR #97, opened 2022-07-07
Move integration tests into a separate module

TomasHofman:
Resolves https://github.com/wildfly-extras/prospero/issues/84.


integration-tests/src/test/java/org/wildfly/prospero/it/cli/InstallTest.java
```diff
@@ -0,0 +1,46 @@
...
+import org.wildfly.prospero.cli.CliMain;
+import org.wildfly.prospero.cli.ReturnCodes;
+import org.wildfly.prospero.cli.commands.CliConstants;
+import org.wildfly.prospero.it.AcceptingConsole;
+import org.wildfly.prospero.test.MetadataTestUtils;
+import picocli.CommandLine;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class InstallTest {
```
spyrkob, line 20: Just a note for future - I think for this sort of tests, we should use distributable generated by dist and run it as external process

TomasHofman: Good idea, I'm going to do that next.

integration-tests/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <skipTests>true</skipTests>
+                    <reuseForks>false</reuseForks> <!-- Don't reuse forks to avoid jboss-modules hanging on in the classpath -->
```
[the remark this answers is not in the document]
TomasHofman: Sure.

TomasHofman:
> What do you think about replacing the integration-tests profile in integration-tests/pom.xml with a profile in root pom.xml that would simply add integration-tests module?

What I like about the current approach is that integration tests will always get recompiled, making it easier to spot issues during API changes. But if you have strong preference for the way you suggested I can change it.

spyrkob:
> What I like about the current approach is that integration tests will always get recompiled, making it easier to spot issues during API changes
That's a good point, let's leave it as it is

spyrkob (approved):

Issue #73, opened 2022-06-28
Class loader issue with jboss-modules


TomasHofman:
```
[thofman@rama prospero]$ dist/build/target/prospero-1.0-SNAPSHOT/bin/prospero.sh install --dir wf-core --provision-config wildfly-core-config.yaml --fpl "wildfly-core@maven(org.jboss.universe:community-universe):19.0"
Listening for transport dt_socket at address: 8787
Installing wildfly-core@maven(org.jboss.universe:community-universe):19.0
Feature-packs resolved.0%
Packages installed. 100%
JBoss modules installed. 100%
Generating configurationError while executing operation 'install': Failed to invoke config generator org.wildfly.galleon.plugin.config.generator.WfConfigGenerator
Jun 28, 2022 12:28:51 PM org.wildfly.prospero.cli.commands.InstallCommand call
ERROR: Error while executing operation 'install': Failed to invoke config generator org.wildfly.galleon.plugin.config.generator.WfConfigGenerator
org.jboss.galleon.ProvisioningException: Failed to invoke config generator org.wildfly.galleon.plugin.config.generator.WfConfigGenerator
	at org.wildfly.galleon.plugin.WfInstallPlugin.generateConfigs(WfInstallPlugin.java:740)
	at org.wildfly.galleon.plugin.WfInstallPlugin.postInstall(WfInstallPlugin.java:471)
	at org.jboss.prospero//org.jboss.galleon.runtime.ProvisioningRuntime$3.visitPlugin(ProvisioningRuntime.java:263)
	at org.jboss.prospero//org.jboss.galleon.runtime.ProvisioningRuntime$3.visitPlugin(ProvisioningRuntime.java:260)
	at org.jboss.prospero//org.jboss.galleon.layout.ProvisioningLayout$Handle.visitPlugins(ProvisioningLayout.java:285)
	at org.jboss.prospero//org.jboss.galleon.layout.ProvisioningLayout.visitPlugins(ProvisioningLayout.java:838)
	at org.jboss.prospero//org.jboss.galleon.runtime.ProvisioningRuntime.provision(ProvisioningRuntime.java:260)
	at org.jboss.prospero//org.jboss.galleon.ProvisioningManager.doProvision(ProvisioningManager.java:644)
	at org.jboss.prospero//org.jboss.galleon.ProvisioningManager.provision(ProvisioningManager.java:403)
[...]

PR #99, opened 2022-07-08
Run CLI integration tests in separate processes

TomasHofman:
This is an idea on how the CLI integration tests could be run in separate processes.

I'm using shaded jar from the CLI module, because the distribution from `dist/build/` module still has the class loader issue (https://github.com/wildfly-extras/prospero/issues/73), but that could be changed later. However, if dist/build/ was to be be used, it needs to be somehow ensured that it's built ("dist" profile should be activated on `-DallTests`?).


spyrkob:
Looks good 👍 

Would it be possible to use prospero script instead of directly calling jar? This way we don't need to read the project version, build the path to cli etc.

> "dist" profile should be activated on -DallTests
+1

TomasHofman:
I updated the PR to use `prospero` script. It simplifies the execution indeed.

TomasHofman:
Rebased

spyrkob:
Thanks!

PR #101, opened 2022-07-08
Externalize known server combinations

### LINKED ISSUE #93: Externalise the configuration of WellKnownFeaturePacks
All enum values in `WellKnownFeaturePacks` should be derived from an external config file.

This will allow to make sure that Prospero either deal with EAP or WildFly but not both. 

One way to externalise these values is with a properties file that list all required settings, with something like:

```
# list the keys of all known feature pack
prospero.feature-packs= x, y, z
# each feature pack defines its own properties
prospero.feature-pack.x.location=...
prospero.feature-pack.x.channel=...
# optional
prospero.feature-pack.x.pakages=...
# optional
prospero.feature-pack.x.repositories=...
```

spyrkob:
Fixes #93 

@TomasHofman could you take a look when you've got a moment


prospero-common/src/main/java/org/wildfly/prospero/api/WellKnownRepositories.java
```diff
@@ -9,7 +9,7 @@
  */
 public enum WellKnownRepositories implements Supplier<RemoteRepository> {
 
-    MRRC(new RemoteRepository.Builder("mrrc", "default", "https://maven.repository.redhat.com/ga/")
+    MRRC(new RemoteRepository.Builder("mrrc", "default", "https://maven.repository.redhat.com/earlyaccess/all/")
```
jmesnil, line 12: Now that the feature pack location are externalised, why do you still need to know these remote repositories?

*If* they stay, I think you should distinguish between `MRRC` (or `MRRC_GA`) and `MRRC_EA`

prospero-common/src/main/resources/prospero-known-combinations.yaml
```diff
@@ -0,0 +1,25 @@
+---
```
jmesnil, line 1: -1, prospero is a community project and is not the place to provide product specific configuration

prospero-common/src/test/resources/prospero-known-combinations.yaml
```diff
@@ -0,0 +1,31 @@
+---
+- name: "no-channel"
+  location: "wildfly@maven(org.jboss.universe:community-universe):current"
+  packages: []
```
jmesnil, line 4: can the `packages` field be optional instead?

jmesnil (changes requested):

prospero-common/src/main/java/org/wildfly/prospero/api/WellKnownRepositories.java
spyrkob: Mostly they're useful in tests, I should just move them to test tree though

prospero-common/src/main/resources/prospero-known-combinations.yaml
```diff
@@ -0,0 +1,8 @@
...
+- name: "wildfly"
```
TomasHofman, line 2: IIUC this is still not "provisionable" without providing some channel via `--channel` argument?

TomasHofman (approved):

prospero-common/src/main/resources/prospero-known-combinations.yaml
spyrkob: Yes the channel has to be provided separately

jmesnil (approved):

Issue #87, opened 2022-07-01
If prospero is run within Wildfly installation, the --dir should not be mandatory


spyrkob:
If no `--dir` is provided, check if current director has `.galleon` and `.installation\prospero-config.xml`

TomasHofman:
Looking at this.

TomasHofman:
Do you mean `.installation\prospero-config.yaml` or `.installation\provisioning.xml` as the second file?

PR #105, opened 2022-07-11
Make --dir command optional, assume "." default value

TomasHofman:
Resolves https://github.com/wildfly-extras/prospero/issues/87.

The `install` command remains as it was - `--dir` is mandatory and must be an empty dir.

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
@@ -1,0 +1,39 @@
+package org.wildfly.prospero.cli;
+
+import org.jboss.galleon.ProvisioningException;
+import org.jboss.logging.Logger;
+import org.wildfly.prospero.actions.Console;
+import org.wildfly.prospero.api.exceptions.OperationException;
+import org.wildfly.prospero.cli.commands.CliConstants;
+import picocli.CommandLine;
+
+/**
+ * Handles exceptions that happen during command executions.
+ */
+public class ExecutionExceptionHandler implements CommandLine.IExecutionExceptionHandler {
+
+    private final Logger logger = Logger.getLogger(this.getClass());
+    private final Console console;
+
+    public ExecutionExceptionHandler(Console console) {
+        this.console = console;
+    }
+
+    @Override
+    public int handleExecutionException(Exception ex, CommandLine commandLine, CommandLine.ParseResult parseResult)
+            throws Exception {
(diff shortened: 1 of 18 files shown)


prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
+package org.wildfly.prospero.cli;
```
spyrkob, line 1: Can you add the license header please?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java
```diff
@@ -23,4 +29,24 @@ public AbstractCommand(Console console, ActionFactory actionFactory) {
         this.console = console;
         this.actionFactory = actionFactory;
     }
+
+    static Path determineInstallationDirectory(Optional<Path> directoryOption) {
```
spyrkob, line 33: do the new methods need to be static?

prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
TomasHofman: +1 will update my idea settings finally and add the headers in a new commit.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java
TomasHofman: Not necessarily, I like to mark the methods static if they are not dependent on the instances' inner state and don't manipulate the inner state, to highlight that fact. But it's matter of taste... Let me know if you still prefer non-static :].

prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
TomasHofman: I created new PR where copyright headers are added in all java files: https://github.com/wildfly-extras/prospero/pull/106

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java
spyrkob: Fair enough. I tend to try and avoid static methods when possible, but like you say it's matter of taste.
Happy to leave that in

PR #107, opened 2022-07-12
Add --version option to the main command

TomasHofman:


prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/MainCommand.java
```diff
@@ -45,4 +51,11 @@ public Integer call() {
         return ReturnCodes.INVALID_ARGUMENTS;
     }
 
+    static class VersionProvider implements CommandLine.IVersionProvider {
+        @Override
+        public String[] getVersion() throws Exception {
+            Manifest manifest = new Manifest(getClass().getClassLoader().getResourceAsStream("META-INF/MANIFEST.MF"));
```
spyrkob, line 57: I think that'll only work if Prospero is distributed as a single shaded jar, not eg. when using jboss-modules

spyrkob (changes requested):

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/MainCommand.java
TomasHofman: That's probably true, I didn't thought of the jboss-modules distribution. I will look into that.

TomasHofman: I updated the PR, the VersionProvider now goes through all the MANIFESTs to locate the prospero-cli manifest and gets the version from there.

spyrkob (approved):
LGTM, thanks!

PR #110, opened 2022-07-14
Channel command

TomasHofman:
Resolves https://github.com/wildfly-extras/prospero/issues/41

This has two commits, one is just refactoring, so it more readable checking them one by one.


prospero-cli/src/main/java/org/wildfly/prospero/cli/CliMain.java
```diff
@@ -72,12 +76,18 @@ public static CommandLine createCommandLine(Console console, ActionFactory actio
...
         commandLine.addSubcommand(new RepositoryCommand(console, actionFactory));
+        commandLine.addSubcommand(new ChannelCommand(console, actionFactory));
 
         CommandLine repoCmd = commandLine.getSubcommands().get(CliConstants.Commands.REPOSITORY);
         repoCmd.addSubcommand(new RepositoryAddCommand(console, actionFactory));
         repoCmd.addSubcommand(new RepositoryRemoveCommand(console, actionFactory));
         repoCmd.addSubcommand(new RepositoryListCommand(console, actionFactory));
 
+        CommandLine channelCmd = commandLine.getSubcommands().get(CliConstants.Commands.CHANNEL);
+        channelCmd.addSubcommand(new ChannelAddCommand(console, actionFactory));
```
spyrkob, line 87: With those small subcommands, do you think it would be tidier if they were inner classes of `ChannelCommand`? Or maybe moved to a separate package per sub-channel?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/ChannelListCommand.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public Integer call() throws Exception {
+        Path installationDirectory = determineInstallationDirectory(directory);
+        MetadataAction metadataAction = actionFactory.metadataActions(installationDirectory);
+        List<ChannelRef> channels = metadataAction.getChannels();
+
+        for (ChannelRef channel: channels) {
+            console.println(String.join(", ", channel.getGavOrUrlString()));
```
spyrkob, line 48: Why do we need the `join`?

prospero-common/src/main/java/org/wildfly/prospero/model/ProsperoConfig.java
```diff
@@ -52,9 +53,36 @@ public List<RepositoryRef> getRepositories() {
     }
 
     public void addChannel(ChannelRef channelRef) {
+        // Check that neither GAV nor URL of added channel is equal to GAVs or URLs of existing channels.
+        if (StringUtils.isNotBlank(channelRef.getGav())) {
```
spyrkob, line 57: Couldn't this be replaced with `channelRef. getGavOrUrlString`? For both new channelRef and existing ones it should give either the gav or url that can be compared straight away?

spyrkob (commented):
A couple questions, otherwise looks good

prospero-cli/src/main/java/org/wildfly/prospero/cli/CliMain.java
TomasHofman: I think both ways would work, they don't contain very much logic... Will update.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/ChannelListCommand.java
TomasHofman: That's a relic, thanks for spotting. I originally wanted to print both url and gav, a then we settled on gav taking precedence.

prospero-common/src/main/java/org/wildfly/prospero/model/ProsperoConfig.java
TomasHofman: So the config containing channels with both GAV and URL are not completely forbidden yet.

I think the problematic case is when adding a URL-containing-channel, and an existing channel from metadata would have both GAV and URL (equal to the one that is to be added) defined. The URL would then be masked if we queried just `getGavOrUrlString()`.

I wanted to prevent repetition of any given GAV or URL in the config. Plus at this level it's not guaranteed that channel to be added also doesn't contain both GAV and URL. Maybe it's not needed to be as strict though.

TomasHofman:
I also changed `MetadataAction#addRepository()` to void return type (rather than boolean), so throws exception if the input cannot be added. It's to keep validation at single place and make it consistent with the other methods in the class.

prospero-common/src/main/java/org/wildfly/prospero/model/ProsperoConfig.java
```diff
@@ -52,26 +53,52 @@ public List<RepositoryRef> getRepositories() {
...
+
     /**
      * Adds a repository to the config. If the repository already exists ({@code id} and {@code url} matches), it is not added.
      * Adding a repository with existing {@code id} but different {@code url} causes {@code IllegalArgumentException}
      * @param repository
      * @throws IllegalArgumentException if a repository with given id but different URL is already present
      * @return true if the repository was added, false if the repository already exists
      */
-    public boolean addRepository(RepositoryRef repository) {
+    public void addRepository(RepositoryRef repository) {
```
spyrkob, line 93: I'm not sure about this. Is adding an existing repository a fault? Especially one that results in unchecked exception? A caller might be able to continue - after the call the repositories are in a state they were supposed to be.

I think a conflicting repository (same id, different url) is a different case- something went wrong and we cannot reasonable handle that.

prospero-common/src/main/java/org/wildfly/prospero/model/ChannelRef.java
```diff
@@ -79,19 +79,23 @@ public static ChannelRef fromString(String gavOrUrl) {
...
-            if (isValidGav(gavOrUrl)) {
+            if (isValidGaOrGav(gavOrUrl)) {
                 return new ChannelRef(gavOrUrl, null);
             }
             throw new IllegalArgumentException(String.format("Given string is not valid Maven GAV or URL: '%s'", gavOrUrl));
         }
     }
 
-    private static boolean isValidGav(String gav) {
+    private static boolean isValidGaOrGav(String gav) {
```
spyrkob, line 89: should we rename Gav too Coordinate?

prospero-common/src/main/java/org/wildfly/prospero/model/ProsperoConfig.java
spyrkob: hmm that's an interesting scenario. We cannot guarantee that a URL based channelRef isn't the same as GAV.

Let's say we have `org.test:channel-one` and user adds `https://maven.repo/org/test/channel-one/1.0.0/channel-one-1.0.0-channel.yaml`. They're both the same channel (assuming there's no newer version), but there's no way for us to tell.

prospero-common/src/main/java/org/wildfly/prospero/model/ProsperoConfig.java
TomasHofman: OK, that makes sense too :). Dropping the commit.

prospero-common/src/main/java/org/wildfly/prospero/model/ChannelRef.java
TomasHofman: Renamed to `isValidCoordinate`.

prospero-common/src/main/java/org/wildfly/prospero/model/ProsperoConfig.java
spyrkob: Let's got with this for now and change the serialisation to allow either GAV or URL. I think that'll solve this issue :)

spyrkob:
@TomasHofman could you rebase please?

TomasHofman:
Sure, rebased.

spyrkob:
Thanks @TomasHofman!

PR #112, opened 2022-07-15
[#111] Remote repositories have update policy set to ALWAYS

### LINKED ISSUE #111: Remote repositories should have ALWAYS update policy
To resolve the latest versions of channels and artifacts, repositories need to have the update policy set to ALWAYS

spyrkob:
Fixes #111


integration-tests/src/test/java/org/wildfly/prospero/it/commonapi/UpdateTest.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static org.junit.Assert.assertEquals;
+
+public class UpdateTest extends WfCoreTestBase {
+
+    private static final String OUTPUT_DIR = "target/server";
```
TomasHofman, line 62: Just a theoretical note / work for future - I've been thinking that by using this hardcoded "target/server" path we block ourselves from running test cases in parallel (at least I think surefire is able to do that, currently it's disabled). But we use this on plenty of other places still. Alternative is TempFolder.

TomasHofman (approved):
Seems OK to me.

integration-tests/src/test/java/org/wildfly/prospero/it/commonapi/UpdateTest.java
spyrkob: Yes I think we could use a temp folder there.

We might run into some issues with simultaneous maven local-cache updates with parallel tests I think, but it'd be interesting to try

PR #114, opened 2022-07-15
Use temp folder instead of hardcoded path to provision servers

spyrkob:
Use TemporaryFolder Rule instead of hardcoded path for provisioning test servers

cc @TomasHofman


integration-tests/src/test/java/org/wildfly/prospero/it/commonapi/InstallationHistoryActionTest.java
```diff
@@ -78,22 +59,19 @@ public void tearDown() throws Exception {
     public void listUpdates() throws Exception {
         // installCore
         provisionConfigFile = MetadataTestUtils.prepareProvisionConfig(CHANNEL_BASE_CORE_19);
-        Path installDir = Paths.get(OUTPUT_PATH.toString());
-        if (Files.exists(installDir)) {
-            throw new ProvisioningException("Installation dir " + installDir + " already exists");
-        }
+        Path installDir = Paths.get(outputPath.toString());
```
TomasHofman, line 62: Line `Path installDir = Paths.get(outputPath.toString());` could be removed?

TomasHofman (approved):

PR #120, opened 2022-07-20
Prospero config no update

### LINKED ISSUE #118: Remove url from prospero-config.yaml if gav is present
`gav` in channel definition takes precedence over `url`. If channel defines both, only `gav` should be persisted

### LINKED ISSUE #119: Don't write prospero-config.yaml on updates
There's no need to write prospero-config on update as the channels and repositories are not changed.

spyrkob:
Fixes #118 
Fixes #119


spyrkob:
@TomasHofman could you take a look when you've got a moment?

TomasHofman (approved):
Looks good

PR #123, opened 2022-07-22
Measure and display installation time

TomasHofman:
I used this to measure performance difference when testing parallel stream for version resolution, but maybe we would like to have that it main?

```
$ ./prospero install --dir tmp/wf-core --provision-config tmp/wildfly-core-config-patterns.yaml --fpl "wildfly-core@maven(org.jboss.universe:community-universe):19.0" 
Installing wildfly-core@maven(org.jboss.universe:community-universe):19.0
Feature-packs resolved.0%
Packages installed. 100%
JBoss modules installed. 1%
Configurations generated.
Installation completed in 36.44 seconds
```


spyrkob:
Could you move it to the `InstallCommand` instead?
It might also be worth adding it to the `UpdateCommand`

TomasHofman:
Updated, also realized the message should be in a resource bundle.

spyrkob (approved):

Issue #100, opened 2022-07-08
Mutual usage of --channel-repo and -provision-config parameters


marekkopecky:
Prospero allows to use --channel-repo and -provision-config together, but can't work with both of them

TomasHofman:
I'm going to take a look.

TomasHofman:
When using `--provision-config`, the `--remote-repositories` are indeed ignored. So either they should be taken into account as well, or this should be somehow documented in the usage message.

I would personally expect the `--remote-repositories` to be functional always. It might be also surprising that `--remote-repositories` overrides repositories predefined in the named feature pack definitions. Maybe we could have `--remote-repositories` that would be additive and `--override-remote-repositories` that would override?

WDYT @spyrkob?

spyrkob:
@TomasHofman do you have a use case in mind where specifying both `--provision-config` and `--remote-repositories` would be useful?

In my mind, user should use either `provisioning-config.yaml` or command line parameters, not mix the two. The config file already lets user specify repositories, so adding/changing them via command line seems unnecessary and potentially a bit confusing.

We could use the picocli exclusive argument group for those two, I think that would fix both the help content and show error if user attempts to use both arguments.

> Maybe we could have --remote-repositories that would be additive and --override-remote-repositories that would override?
I feel like we already have too many command line options :) My thinking was that using `--remote-repositories` and `--channel` should be equivalent to a `provisioning-config.yaml` with the same values.

PR #125, opened 2022-07-25
Mark and validate --provision-config and --channel as alternatives

TomasHofman:
Update usage message and throw an error when the two options are used together (currently --channel silently ignored).

Resolves https://github.com/wildfly-extras/prospero/issues/100


spyrkob:
Should the same apply to `--remote-repositories` and `--provision-config`?

TomasHofman:
OK. I will also try to put more detailed examples into the usage message.

TomasHofman:
Screenshot from 2022-07-25 10-36-13

TomasHofman:
Should `--channel` also support GAV?

TomasHofman:
I find the examples are wrong, `prospero install --fpl=<GA-or-URL> --provision-config=<path> [OPTION]...` doesn't work.

Only these are valid:

```
prospero install --fpl=<predefined-name> [--channel=<URL> --remote-repositories=<URL>[,...]] [OPTION]...
prospero install --fpl=<GA-or-URL> --provision-config=<path> [OPTION]...
prospero install --definition=<path> [OPTION]...
```

spyrkob:
> Should --channel also support GAV?
Yes, I think we can add GAV support

TomasHofman:
Updated.

spyrkob (approved):

PR #180, opened 2022-11-08
Resolve #168 Add README file to .installation directory warning the files should not be edited

soul2zimate:
Resolve #168 Add README file to .installation directory warning the files should not be edited

PR #182, opened 2022-11-09
Fix java.lang.ClassCastException

soul2zimate:
```
java.lang.ClassCastException: class com.fasterxml.jackson.databind.exc.MismatchedInputException cannot be cast to class org.wildfly.channel.UnresolvedMavenArtifactException (com.fasterxml.jackson.databind.exc.MismatchedInputException and org.wildfly.channel.UnresolvedMavenArtifactException are in unnamed module of loader 'app')
	at org.wildfly.prospero.api.exceptions.ArtifactResolutionException.failedArtifacts(ArtifactResolutionException.java:81)
	at org.wildfly.prospero.cli.ExecutionExceptionHandler.handleExecutionException(ExecutionExceptionHandler.java:64)
	at picocli.CommandLine.execute(CommandLine.java:2088)
	at org.wildfly.prospero.cli.CliMain.main(CliMain.java:54)
```

Fix `java.lang.ClassCastException`. It's not necessarily an `UnresolvedMavenArtifactException` e.g.  The exception has been thrown from https://github.com/wildfly-extras/prospero/blob/1.0.0.Beta4/prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java#L86

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/Messages.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/Messages.java
@@ -25,6 +25,7 @@ import org.jboss.logging.annotations.Cause;
 import org.jboss.logging.annotations.Message;
 import org.jboss.logging.annotations.MessageBundle;
 import org.wildfly.channel.InvalidChannelException;
+import org.wildfly.channel.UnresolvedMavenArtifactException;
 import org.wildfly.prospero.api.exceptions.ArtifactResolutionException;
 import org.wildfly.prospero.api.exceptions.ChannelDefinitionException;
 import org.wildfly.prospero.api.exceptions.MetadataException;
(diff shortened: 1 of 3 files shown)


spyrkob:
thanks @soul2zimate.

I think https://github.com/wildfly-extras/prospero/blob/1.0.0.Beta4/prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java#L86 should throw `MetadataException` instead. We should change the `ArtifactResolutionException` to only accept `UnresolvedMavenArtifactException` (or a list of `ArtifactCoordinates`). WDYT?

soul2zimate:
oh I didn't see there is `MetadataException`, it does seem it's the correct one to throw in that case.

spyrkob (approved):
Thanks @soul2zimate

spyrkob:
@soul2zimate could you rebase this PR against `future` branch please?

soul2zimate:
rebased to future branch and resolve all the conflicts. It doesn't compile even I build a local `installation-manager-api`. Then I see future branch waits for Channel and Manifest split.

Also https://github.com/wildfly-extras/prospero/pull/180 is in main, but it's not in future, Do you need a PR to future for that as well ?

PR #183, opened 2022-11-10
Add support for defining known configurations in provisioning.xml

### LINKED ISSUE #140: Add support for defining known configurations in provisioning.xml
To specify more complex combinations, we should be able to package provisioning.xml within prospero and use something like:

```
- name: "eap-8.0-beta"
  galleonConfiguration: "galleon/eap-8.0-provisioning.xml"
  channelGavs:
  - "org.jboss.eap.channels:eap-8.0-beta"
  repositories:
  - id: "mrrc"
    url: "https://maven.repository.redhat.com/earlyaccess/all/" 
```

TomasHofman:
Resolves #140

### DIFF
--- a/prospero-cli/src/test/resources/galleon-provisioning.xml
+++ b/prospero-cli/src/test/resources/galleon-provisioning.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0" ?>
+
+<!--
+  ~ Copyright 2022 Red Hat, Inc. and/or its affiliates
+  ~ and other contributors as indicated by the @author tags.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<installation xmlns="urn:jboss:galleon:provisioning:3.0">
+    <feature-pack location="org.wildfly.core:wildfly-core-galleon-pack@maven(org.jboss.universe:community-universe)#latest">
+        <packages>
+            <include name="package1"/>
+            <exclude name="package2"/>
+        </packages>
+    </feature-pack>
+    <options>
+        <option name="jboss-bulk-resolve-artifacts" value="true"/>
+    </options>
+</installation>
--- a/prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
@@ -18,6 +18,7 @@
 package org.wildfly.prospero.api;
 
 import java.io.IOException;
+import java.net.URI;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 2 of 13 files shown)


TomasHofman:
Going to check how this merges on the future branch.

spyrkob:
I think the known-feature-pack should either have `galleonConfiguration` or `location`/`packages` fields, never both. Otherwise we have to think about merging those configs.
Maybe we can even get rid of the latter and only use `galleonConfiguration`, WDYT?

TomasHofman:
No, I think it would be better to check `galleonConfiguration` and `location`/`packages` are exclusive in the `KnownFeaturePack` class. That would throw an exception during parsing hopefully.

spyrkob:
> IMO it would be nice if prospero offered an ability to install with only inline parameters, without having to have any files around. That would imply the need to keep the location/packages parameters I think.

The insides of known-feature-pack shouldn't be visible to the user, this is purely internal configuration.
To install with only parameters a user would do: `prospero install --fpl=<FP_GAV> --remote-repositories=<REPO_URLs> --manifest=<MANIFEST> --dir server`

TomasHofman:
You are right again, I lost the context. I think it's OK to only support one way of configuring this.

TomasHofman:
I rebased this on the "future" branch. Removed references to `location` and `packages` from the `KnownFeaturePacks`.

There is one remaining thing that I think needs to be solved, and that is how to reference the `*-provisioning.xml` files from the `prospero-known-combinations.yaml`, in cases when:

1) the files are packaged on the classpath,
2) and when the files are present on the file-system (also what should a current dir be, from which to resolve relative paths).

File system case:

TomasHofman:
Or do we only want to support the classpath case? (Currently it is. It's about how the manager is to be distributed...)

In that case the "classpath:" prefix is not needed and I can just always load the XMLs via `getClassLoader().getResource(...)`.

spyrkob:
I think the classpath option is enough. We only load the known-feature-packs.yaml via classpath as well

prospero-cli/src/test/resources/galleon-provisioning.xml
```diff
@@ -0,0 +1,30 @@
+<?xml version="1.0" ?>
+
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one or more
```
spyrkob, line 4: can you change the license header to https://github.com/wildfly-extras/prospero/blob/main/prospero-common/src/main/java/org/wildfly/prospero/actions/Console.java#L1-L16 ?

prospero-common/src/main/resources/prospero-known-combinations.yaml
```diff
@@ -1,7 +1,6 @@
 ---
 - name: "wildfly"
-  location: "org.wildfly:wildfly-galleon-pack"
-  packages: []
+  galleonConfiguration: "classpath:wildfly-provisioning.xml"
```
spyrkob, line 3: Should the `wildfly-provisioning.xml` be part of the PR?

TomasHofman: It should be :).

TomasHofman:
I updated the PR.

There is now default "production" `provisioning.xml` for Wildfly, which uses FPL "wildfly@maven(org.jboss.universe:community-universe):current#latest".

In the end, I also had to implement "classpath:" scheme for `*-provisioning.xml` files. It's true that when working with predefined FPLs, they are currently always on classpath, but the logic further on (`ProvisioningDefinition#toProvisioningConfig()`) needs to be able to handle both filesystem references and classpath references - it's common code for both use cases.

Separate thought: does it make sense to have Wildfly as predefined FPL, when there is no publicly available channel for Wildfly? We don't allow installation without channel... Maybe it gets replaced by EAP8 in future?

spyrkob (commented):
Looks good, last couple of nitpicks

spyrkob:
>Separate thought: does it make sense to have Wildfly as predefined FPL, when there is no publicly available channel for Wildfly? We don't allow installation without channel... Maybe it gets replaced by EAP8 in future?

The idea was to make it easier by hiding the Wildfly FPL, but yes without the channel it's not very useful.

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -156,32 +157,34 @@ public List<Channel> getChannels() {
...
         return definition;
     }
 
     public ProsperoConfig getProsperoConfig() {
         return new ProsperoConfig(channels);
     }
 
-    public ProvisioningConfig toProvisioningConfig() throws ProvisioningException {
-        if (getFpl() != null) {
+    public ProvisioningConfig toProvisioningConfig() throws ProvisioningException, XMLStreamException {
```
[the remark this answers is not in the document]
TomasHofman: Updated...

spyrkob (approved):

spyrkob:
Thanks @TomasHofman!

PR #184, opened 2022-11-10
Unify remote-repositories parameter

spyrkob:
`--repositories` overrides any repositories already configured either in installed server, or known-feature-packs.
In operations on installed server (update, revert) the repositories defined that way are not persisted. 
The values permitted are either comma-separated URLs or pairs or ID::URL

Renamed `--local-repo` to `--local-cache` and `--channel` to `--manifest`

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/api/TemporaryRepositoriesHandler.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/api/TemporaryRepositoriesHandler.java
@@ -19,10 +19,7 @@ package org.wildfly.prospero.api;
 
 import org.wildfly.channel.Channel;
 import org.wildfly.channel.Repository;
-import org.wildfly.prospero.Messages;
 
-import java.net.MalformedURLException;
-import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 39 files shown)


spyrkob:
@TomasHofman could you take a look, do the CLI changes make sense to you?

prospero-common/src/main/java/org/wildfly/prospero/api/TemporaryRepositoriesHandler.java
```diff
@@ -0,0 +1,52 @@
...
+            final List<Repository> repositories = mergeRepositories(oc.getRepositories(), additionalRepositories);
+            final Channel c = new Channel(oc.getSchemaVersion(), oc.getName(), oc.getDescription(), oc.getVendor(),
+                    oc.getChannelRequirements(), repositories, oc.getManifestRef());
+            mergedChannels.add(c);
+        }
+
+        return mergedChannels;
+    }
+
+    private static List<Repository> mergeRepositories(List<Repository> originalRepositories, List<Repository> additionalRepositories) {
```
TomasHofman, line 49: Should it be called `mergeRepositories` when it overrides them? Any plans for this behavior to be modifiable?

TomasHofman:
Seems OK to me after a quick read. I think it's a good change.

prospero-common/src/main/java/org/wildfly/prospero/api/TemporaryRepositoriesHandler.java
spyrkob: Good catch, that's leftover from some experiments. I'll remove it

Issue #152, opened 2022-09-16
When using a custom fpl, it should be possible to specify channel and repositories using command parameters only.


spyrkob:
Currently when using a custom FPL, user has to provide a provision-config file with channels and repositories, eg:
```
./prospero.sh --dir=test --fpl=org.wildfly:wildfly-ee-galleon-pack --provision-config=wfly-config.yaml
```
where wfly-config.yaml contains:
```
channels:
  - url: "file:///path/to/repository"
repositories:
  - id: "central"
    url: "https://repo1.maven.org/maven2/"
  - id: "jboss-public"
    url: "https://repository.jboss.org/nexus/content/groups/public/"
```

Instead it should be possible to pass those via CLI paramaters like:
```
./prospero.sh --dir=test \
    --fpl=org.wildfly:wildfly-ee-galleon-pack \
    --channel=/path/to/channel.yaml\
    --remote-repositories=https://repo1.maven.org/maven2/,https://repository.jboss.org/nexus/content/groups/public/
```

TomasHofman:
I'm going to look at this.

PR #198, opened 2022-11-18
Rename --provision-config parameter to --channels

TomasHofman:
...and allow combination with --repositories

Resolves https://github.com/wildfly-extras/prospero/issues/152

Work in progress, do not merge yet.


examples/wildfly-core-channel.yaml
```diff
@@ -0,0 +1,10 @@
+---
+schemaVersion: "2.0.0"
+repositories:
+  - id: "central"
+    url: "https://repo1.maven.org/maven2/"
+  - id: "jboss-public"
+    url: "https://repository.jboss.org/nexus/content/groups/public/"
+manifest:
+  url: "file:///home/thofman/Projects/prospero/examples/wildfly-core-manifest.yaml"
```
spyrkob, line 9: I guess the github url will be in the final PR?

prospero-common/src/main/java/org/wildfly/prospero/actions/ProvisioningAction.java
```diff
@@ -57,26 +60,23 @@ public ProvisioningAction(Path installDir, MavenSessionManager mavenSessionManag
...
-     * @param config
-     * @param channels
-     * @throws ProvisioningException
-     * @throws MetadataException
+     * @param provisioningDefinition prospero provisioning definition
      */
-    public void provision(ProvisioningConfig config, List<Channel> channels) throws ProvisioningException, OperationException {
-        doProvision((pm,options)->pm.provision(config, options), enforceChannelNames(channels));
-    }
+    public void provision(ProvisioningDefinition provisioningDefinition)
```
spyrkob, line 67: I think I'd rather keep this using ProvisioningConfig and channels directly rather then ProvisioningDefinition

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactUtils.java
```diff
@@ -31,25 +34,39 @@ public static int compareVersion(Artifact first, Artifact other) {
...
+                    } else if (gav.length == 3) {
+                        return clazz.getDeclaredConstructor(String.class, String.class, String.class)
+                                .newInstance(gav[0], gav[1], gav[2]);
+                    }
                 }
-            } else {
                 // assume the string is a path
                 try {
-                    return new ChannelManifestCoordinate(Paths.get(urlGavOrPath).toAbsolutePath().toUri().toURL());
+                    return clazz.getDeclaredConstructor(URL.class)
```
spyrkob, line 62: Maybe add a check that the file exists?

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactUtils.java
```diff
@@ -31,25 +34,39 @@ public static int compareVersion(Artifact first, Artifact other) {
         return new ComparableVersion(first.getVersion()).compareTo(new ComparableVersion(other.getVersion()));
     }
 
-    public static ChannelManifestCoordinate manifestFromString(String urlGavOrPath) {
+    public static ChannelManifestCoordinate manifestCoordFromString(String urlGavOrPath) {
+        return coordinateFromString(urlGavOrPath, ChannelManifestCoordinate.class);
+    }
+
+    public static ChannelCoordinate channelCoordFromString(String urlGavOrPath) {
```
spyrkob, line 41: can you add tests for both `channelCoordFromString` and `manifestCoordFromString` - want to make sure we catch if the constructors change

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -51,47 +54,73 @@ public class ProvisioningDefinition {
 
     public static final RepositoryPolicy DEFAULT_REPOSITORY_POLICY = new RepositoryPolicy(true, RepositoryPolicy.UPDATE_POLICY_ALWAYS, RepositoryPolicy.CHECKSUM_POLICY_FAIL);
 
+    /**
+     * Galleon feature pack location. Can be either a well-known name (like "eap-8.0") that references a predefined combination
```
spyrkob, line 58: eap-8.0/wildfly

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -51,47 +54,73 @@ public class ProvisioningDefinition {
...
+     * of Galleon configuration and channels, or a standard Galleon feature pack location string (see
+     * https://docs.wildfly.org/galleon/#_feature_pack_location).
+     */
     private final String fpl;
+
+    /**
+     * This field would only contain channels parsed from the prosper-known-combinations.yaml file ("well-known feature packs"),
```
spyrkob, line 65: The comments talk about command line attributes, while the `ProvisioningDefinition` is in prospero-common. I think the ProvisioningDefinition should be moved to cli module

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -104,23 +133,29 @@ private boolean channelsMissingManifest() {
...
+                new RemoteRepository.Builder("MRRC", null, "https://maven.repository.redhat.com/ga/").build(),
+                new RemoteRepository.Builder("central", null, "https://repo1.maven.org/maven2").build()
+        );
+    }
+
+    private static List<Repository> extractRepositoriesFromChannels(List<Channel> channels) {
+        return channels.stream().flatMap(c -> c.getRepositories().stream()).collect(Collectors.toList());
+    }
+
+    private void setKnownFeaturePackChannels(List<Repository> overrideRepositories,
```
spyrkob, line 148: what happens if channelCoordinates is not empty?

Do you think this method could return List<Channel> and assign this.channels in the constructor?

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -104,23 +133,29 @@ private boolean channelsMissingManifest() {
         return channels.stream().anyMatch(c->c.getManifestRef() == null);
     }
 
-    private void setUpBuildEnv(List<Repository> additionalRepositories, Optional<Path> provisionConfigFile,
-                               Optional<ChannelManifestCoordinate> manifestRef, List<Channel> channels) throws IOException {
-        if (!provisionConfigFile.isPresent() && !manifestRef.isPresent()) {
-            if (!additionalRepositories.isEmpty()) {
-                this.channels = TemporaryRepositoriesHandler.addRepositories(channels, additionalRepositories);
+    // TODO: externalize
+    private static List<RemoteRepository> defaultChannelResolutionRepositories() {
```
spyrkob, line 137: How about overriding (or at least supplementing) those with `overrideRepositories`?

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -164,29 +203,48 @@ public ProvisioningConfig toProvisioningConfig() throws ProvisioningException {
...
+     *
+     * @param versionResolverFactory a VersionResolverFactory instance to perform the channel resolution
+     * @return Channel instances
+     */
+    public List<Channel> resolveChannels(VersionResolverFactory versionResolverFactory) {
+        try {
+            List<Channel> channels = new ArrayList<>(this.channels);
+
+            if (!channelCoordinates.isEmpty()) {
+                channels.addAll(versionResolverFactory.resolveChannels(channelCoordinates, defaultChannelResolutionRepositories()));
```
spyrkob, line 217: shouldn't channelCoordinates allow path as well as URL and GAV?

examples/wildfly-core-channel.yaml
TomasHofman: Yes, I think the URL to GitHub are best for this use cases because they are absolute and will stay active. Ideally it should also be pointing to the "main" branch.

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
TomasHofman: I will just change the attribute references to builder methods. I think this class is useful as part of programmatic API in common.

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactUtils.java
TomasHofman: Sure, test added.

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactUtils.java
TomasHofman: I wouldn't do that here. We don't check URLs and GAVs for existence either.

prospero-common/src/main/java/org/wildfly/prospero/actions/ProvisioningAction.java
TomasHofman: OK, I wanted to do channel resolution inside the `ProvisioningAction` class, but it turns out it's OK to do it in the `InstallCommand` as well.

examples/wildfly-core-channel.yaml
TomasHofman: Updated to GH urls.

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
TomasHofman: I was planning originally to have separate CLI option to control this, but attaching this to `--repositories` as well seems OK.

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
TomasHofman: If they were Paths, they were already converted to URLs at this point. The conversion is done in the `ProvisioningDefinition.Builder#setChannelCoordinates()`.

spyrkob: ah right, missed that 👍

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
```diff
@@ -107,20 +111,20 @@ public void errorIfChannelsIsNotPresentAndUsingCustomFplOnInstall() {
                 CliConstants.FPL, "foo:bar");
         assertEquals(ReturnCodes.INVALID_ARGUMENTS, exitCode);
         assertTrue("output: " + getErrorOutput(), getErrorOutput().contains(String.format(
-                CliMessages.MESSAGES.prosperoConfigMandatoryWhenCustomFpl().getMessage(), CliConstants.PROVISION_CONFIG)));
+                CliMessages.MESSAGES.channelsMandatoryWhenCustomFpl().getMessage(), CliConstants.CHANNELS)));
```
spyrkob, line 114: String.format is not needed, the new message doesn't take a parameter

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
```diff
@@ -220,30 +224,33 @@ public void fplAndDefinitionAreNotAllowedTogether() throws Exception {
...
-                .exclusiveOptions(CliConstants.PROVISION_CONFIG, CliConstants.CHANNEL_MANIFEST).getMessage()));
+                .exclusiveOptions(CliConstants.CHANNELS, CliConstants.CHANNEL_MANIFEST).getMessage()));
     }
 
     @Test
-    public void provisionConfigAndRemoteRepoSet() throws IOException {
-        final File provisionConfigFile = temporaryFolder.newFile();
+    public void provisionConfigAndRemoteRepoSet() throws Exception {
+        Path channelsFile = temporaryFolder.newFile().toPath();
+        MetadataTestUtils.prepareChannel(channelsFile, List.of(getClass().getClassLoader().getResource("manifest.yaml")));
```
spyrkob, line 242: maybe instead of adding the test manifest, just mock the manifest file here? just empty temporary file should do I think

prospero-common/src/main/java/org/wildfly/prospero/actions/ProvisioningAction.java
```diff
@@ -57,26 +57,22 @@ public ProvisioningAction(Path installDir, MavenSessionManager mavenSessionManag
...
     /**
-     * Installs feature pack defined by {@code fpl} in {@code installDir}. If {@code fpl} doesn't include version,
-     * the newest available version will be used.
+     * Provision installation according to given ProvisioningDefinition.
      *
-     * @param config
-     * @param channels
-     * @throws ProvisioningException
-     * @throws MetadataException
+     * @param provisioningDefinition prospero provisioning definition
```
spyrkob, line 62: wrong `@param`

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -137,6 +143,10 @@ public ProsperoConfig getProsperoConfig() {
         return new ProsperoConfig(channels);
     }
 
+    public List<Repository> getOverrideRepositories() {
```
spyrkob, line 146: can we remove this? I don't think it's used anywhere

prospero-common/src/main/java/org/wildfly/prospero/api/RepositoryUtils.java
```diff
@@ -18,18 +18,21 @@
...
-
 public class RepositoryUtils {
     public static Repository toChannelRepository(RemoteRepository r) {
         return new Repository(r.getId(), r.getUrl());
     }
 
     public static RemoteRepository toRemoteRepository(String id, String url) {
         return new RemoteRepository.Builder(id, "default", url)
-                .setPolicy(DEFAULT_REPOSITORY_POLICY)
+                .setPolicy(new RepositoryPolicy(true, RepositoryPolicy.UPDATE_POLICY_ALWAYS, RepositoryPolicy.CHECKSUM_POLICY_IGNORE))
```
spyrkob, line 31: Why the `CHECKSUM_POLICY_IGNORE` policy? don't think we should ignore checksum errors

spyrkob (changes requested):

integration-tests/src/test/java/org/wildfly/prospero/it/featurepacks/WildflyFpTest.java
```diff
@@ -66,10 +66,12 @@ public void testInstallProsperoWithWildfly() throws Exception {
         final Path installedProspero = targetDir.resolve("bin").resolve(ExecutionUtils.isWindows()?"prospero.bat":"prospero.sh");
         assertTrue(Files.exists(installedProspero));
 
-        ExecutionUtils.prosperoExecution(CliConstants.Commands.UPDATE, CliConstants.DRY_RUN,
+        // TODO: This doesn't work due to incompatible wildfly-channel library version. Re-enable after we get back to an
+        //  official wildfly-channel release.
+        /*ExecutionUtils.prosperoExecution(CliConstants.Commands.UPDATE, CliConstants.DRY_RUN,
```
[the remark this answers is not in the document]
TomasHofman: It was something like "Class ChannelManifest not found". I think I used an old version of `prospero-manifest.yaml` - it only updates when running the build with `-DallTests`, which I forgot to be doing. It works now.

integration-tests/src/test/resources/manifests/prospero-manifest.yaml
```diff
@@ -0,0 +1,166 @@
+---
```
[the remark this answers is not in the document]
TomasHofman: Done.

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
```diff
@@ -178,6 +160,7 @@ public ProvisioningConfig toProvisioningConfig() throws MetadataException, Provi
             try {
                 return GalleonUtils.loadProvisioningConfig(definition);
             } catch (XMLStreamException e) {
+                e.printStackTrace();
```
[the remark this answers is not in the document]
TomasHofman: Yes, forgot to remove after some debugging.

examples/README.md
```diff
@@ -0,0 +1,87 @@
...
+  --manifest examples/wildfly-27.0.0.Alpha2-manifest.yaml \
+  --dir installation-dir
+```
+
+(Note: the `--manifest ...` option would be optional for predefined "eap-*" options.)
+
+## Installation Referencing a Channel
+
+```shell
+./prospero install --fpl org.wildfly.core:wildfly-core-galleon-pack:20.0.0.Beta2 \
```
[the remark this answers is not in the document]
TomasHofman: If I remove the version, I think the example is going to stop working in the future (when the core-galleon-pack does changes that would not be reflected by the wf-core manifest versioned in this directory).

Maybe I could stress that this readme is intended to developers, not users? Maybe also point out that the FP version would normally be omitted?

BTW my goal was to have some inputs for immediate testing / demonstration, as there are no available manifests around yet and I always have to go and generate a fresh manifest from Wildfly. (I wanted to use wf-core because it's smaller / faster.)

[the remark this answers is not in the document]
TomasHofman: I should fix the 20.0.0.Beta3-SNAPSHOT manifest to 20.0.0.Beta2.

examples/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Description:
+ * The `--fpl` option defines a feature pack to be installed.
+ * The `--channels` option defines a Wildfly Channel (a *Wildfly Channel* references a *Wildfly Channel Manifest* and
+Maven repositories). 
+
+## Installation Referencing a Channel Manifest and Maven Repositories
+
+```shell
+./prospero install --fpl org.wildfly.core:wildfly-core-galleon-pack:20.0.0.Beta2 \
```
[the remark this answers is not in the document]
TomasHofman: OK, I wanted to make this combination valid and then didn't.

Are you OK with making the `--manifest` option valid for all input options (pre-defined fpl, custom fpl, galleon config)? Originally it was valid for pre-defined FPL. I think it would be more transparent behavior.

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
TomasHofman: You are right!

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
TomasHofman: I can give a non-existing URL as a manifest reference, but the channel file need to exists.

prospero-common/src/main/java/org/wildfly/prospero/api/ProvisioningDefinition.java
TomasHofman: Yes.

prospero-common/src/main/java/org/wildfly/prospero/api/RepositoryUtils.java
TomasHofman: Thanks for spotting, another forgotten bit. I was testing resolving manifests from local repo, that didn't have checksums...

examples/README.md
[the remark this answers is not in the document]
TomasHofman: How does missing `org.jboss.invocation:jboss-invocation-jakarta` manifest? I don't see that being a wildfly-core dependency.

examples/README.md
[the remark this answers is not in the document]
spyrkob: Yes that sounds good @TomasHofman 👍

examples/README.md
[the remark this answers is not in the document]
spyrkob: > If I remove the version, I think the example is going to stop working in the future (when the core-galleon-pack does changes that would not be reflected by the wf-core manifest versioned in this directory).

Not sure I follow. Without the version it will install the FP version defined by the manifest - so that will always be the 20.0.0.Beta2 and shouldn't change. Or am I missing something?

> BTW my goal was to have some inputs for immediate testing / demonstration, as there are no available manifests around yet and I always have to go and generate a fresh manifest from Wildfly. (I wanted to use wf-core because it's smaller / faster.)
👍 

> How does missing org.jboss.invocation:jboss-invocation-jakarta manifest? I don't see that being a wildfly-core dependency.

When I changed the core versions to 20.0.0.Beta2 provisioning failed with missing artifact. Maybe it was removed in Beta3

spyrkob (approved):

TomasHofman:
Seems like CI is finally passing :+1:

spyrkob:
Thanks @TomasHofman !

PR #200, opened 2022-11-23
[#186] Display channel diffs

### LINKED ISSUE #186: Show differences for CONFIG_CHANGE history changes
`history --revision` operation currently only shows differences in installed artifacts

For changes in configuration (added/removed channels) it should show output like:
```
Added channels:
  * <channel_name>
     repositories: <id>::<url>,...
     manifest: <url-or-gav>
Removed channels:
   * <channel_name>
      repositories: <id>::<url>,...
      manifest: <url-or-gav>
```

spyrkob:
Fixes #186

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactChange.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactChange.java
@@ -20,46 +20,45 @@ package org.wildfly.prospero.api;
 import org.eclipse.aether.artifact.Artifact;
 import org.wildfly.channel.version.VersionMatcher;
 
+import java.util.Objects;
 import java.util.Optional;
 
-public class ArtifactChange {
-    private Artifact oldVersion;
-    private Artifact newVersion;
-
-    public ArtifactChange(Artifact oldVersion, Artifact newVersion) {
-        this.oldVersion = oldVersion;
-        this.newVersion = newVersion;
+public class ArtifactChange extends Diff {
+    public static ArtifactChange added(Artifact newVersion) {
+        Objects.requireNonNull(newVersion);
+        return new ArtifactChange(toGav(newVersion), null, newVersion.getVersion());
     }
 
+    public static ArtifactChange removed(Artifact oldVersion) {
+        Objects.requireNonNull(oldVersion);
+        return new ArtifactChange(toGav(oldVersion), oldVersion.getVersion(), null);
+    }
+
+    public static ArtifactChange updated(Artifact oldVersion, Artifact newVersion) {
+        Objects.requireNonNull(oldVersion);
+        Objects.requireNonNull(newVersion);
+        return new ArtifactChange(toGav(oldVersion), oldVersion.getVersion(), newVersion.getVersion());
+    }
+
+    private ArtifactChange(String gav, String oldVersion, String newVersion) {
+        super(gav, oldVersion, newVersion);
+    }
+
+    @SuppressWarnings("OptionalGetWithoutIsPresent")
     public String getArtifactName() {
-        if (oldVersion == null) {
-            return toGav(newVersion);
-        } else {
-            return toGav(oldVersion);
-        }
+        // the name has to be present, as it's either old or new artifact's gav
+        return getName().get();
     }
 
     public Optional<String> getOldVersion() {
-        return oldVersion==null?Optional.empty():Optional.of(oldVersion.getVersion());
+        return getOldValue();
     }
 
     public Optional<String> getNewVersion() {
-        return newVersion==null?Optional.empty():Optional.of(newVersion.getVersion());
+        return getNewValue();
     }
 
-    @Override
-    public String toString() {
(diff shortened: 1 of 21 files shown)


spyrkob:
@TomasHofman if you've got a moment could you take a look?

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactChange.java
```diff
@@ -22,44 +22,25 @@
...
+        super(toGav(oldVersion == null?newVersion:oldVersion), oldVersion==null?null:oldVersion.getVersion(), newVersion==null?null:newVersion.getVersion());
     }
 
     public String getArtifactName() {
-        if (oldVersion == null) {
-            return toGav(newVersion);
-        } else {
-            return toGav(oldVersion);
-        }
+        return getName().get();
```
TomasHofman, line 32: Should the `getArtifactName()` return Optional too (same as the Diff)? Or should the name be non-nullable? To avoid a possible NPE inside this method...

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactChange.java
```diff
@@ -22,44 +22,25 @@
...
-        }
-        final String gac = toGav(oldVersion);
-        return String.format("Update [%s]:\t\t %s ==> %s", gac, oldVersion.getVersion(), newVersion.getVersion());
+        return getNewValue();
     }
 
-    private String toGav(Artifact artifact) {
+    private static String toGav(Artifact artifact) {
         final String gac;
         if (artifact.getClassifier() == null || artifact.getClassifier().isEmpty()) {
```
TomasHofman, line 45: I'm wondering why the classifier is part of this string? (Rather then e.g. version.)

TomasHofman:
Example output:

```
[thofman@rama prospero]$ ./prospero history --dir wfhist --revision 390de405
Updates:
  [Updated] org.jboss.universe.producer:wildfly-producers:		1.3.2.Final ==> 1.3.1.Final
  [Updated] com.carrotsearch:hppc:		0.8.1 ==> 0.8.0

Configuration changes:
  [Added] wfcore:
    manifest:		[] ==> file:/home/thofman/tmp/wildfly-core-manifest.yaml
    repositories: 
      [] ==> temp-repo-0::https://repo1.maven.org/maven2/
```

What about adding nouns to explain what the reported items are? E.g. instead of `[Added]` / `[Updated]` there could be  `[Updated artifact]` or `[Added channel]`. I think it may not be clear what the configuration change represent currently.

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactChange.java
spyrkob: For the ArtifactChange, the name can never be null since it's based on the old and new Artifacts which have to include groupId and artifactId

spyrkob:
> What about adding nouns to explain what the reported items are? E.g. instead of [Added] / [Updated] there could be [Updated artifact] or [Added channel]. I think it may not be clear what the configuration change represent currently.

I like that 👍  let me see if I can add it

TomasHofman:
Observation, if I modify the `installer-config.yaml`, I cause an un-versioned change (`prospero history --revision <latest hash>` would report my change). This change will remain un-versioned until I do some CLI operation that modifies the config too (like adding / removing a channel).

Should there be a CLI command that would version my manual changes?

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactChange.java
TomasHofman: Right, didn't realize that. We could add something like `assert getName().isPresent() : "Artifact name should never be null.";` to make that clear for IDE inspections.

prospero-common/src/main/java/org/wildfly/prospero/api/ArtifactChange.java
```diff
@@ -22,44 +22,25 @@
...
 
-public class ArtifactChange {
-    private Artifact oldVersion;
-    private Artifact newVersion;
+public class ArtifactChange extends Diff {
 
     public ArtifactChange(Artifact oldVersion, Artifact newVersion) {
-        this.oldVersion = oldVersion;
-        this.newVersion = newVersion;
+        super(toGav(oldVersion == null?newVersion:oldVersion), oldVersion==null?null:oldVersion.getVersion(), newVersion==null?null:newVersion.getVersion());
```
TomasHofman, line 28: A detail - I like the approach in the ChannelChange, where the instances are created via static constructor methods. It would simplify this constructor code and as well make it clearer how to use the class.

spyrkob:
> Should there be a CLI command that would version my manual changes?

hm I'm not sure. At the moment the idea is that installation-config.yaml changes should only be done via the CLI/API operations.

TomasHofman (approved):
I think it's working nicely!

PR #202, opened 2022-11-25
[Issue-187] Add export/import operation to CLI

### LINKED ISSUE #187: Add export/import operation to CLI
Expose `InstallationExportAction#export` and `InstallationRestoreAction#restore` via CLi operation `prospero clone export --dir=server --path=output.zip` `prospero clone import --dir=server --path=output.zip`

The `import` operation should support maven parameters as well:
- --repositories
- --ofline
- --local-cache
- --no-resolve-local-cache

gaol:
Fixes #187 

@spyrkob would you please review? thanks

### DIFF
--- a/integration-tests/src/test/java/org/wildfly/prospero/it/cli/CloneTest.java
+++ b/integration-tests/src/test/java/org/wildfly/prospero/it/cli/CloneTest.java
@@ -1,0 +1,183 @@
+/*
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.wildfly.prospero.it.cli;
+
+import org.apache.commons.io.FileUtils;
+import org.jboss.galleon.config.FeaturePackConfig;
+import org.jboss.galleon.config.ProvisioningConfig;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.wildfly.channel.Channel;
+import org.wildfly.channel.ChannelManifest;
+import org.wildfly.channel.ChannelManifestMapper;
+import org.wildfly.channel.Repository;
+import org.wildfly.channel.Stream;
+import org.wildfly.prospero.actions.InstallationExportAction;
+import org.wildfly.prospero.api.InstallationMetadata;
+import org.wildfly.prospero.cli.ReturnCodes;
+import org.wildfly.prospero.cli.commands.CliConstants;
+import org.wildfly.prospero.it.ExecutionUtils;
+import org.wildfly.prospero.it.commonapi.WfCoreTestBase;
+import org.wildfly.prospero.model.ManifestYamlSupport;
+import org.wildfly.prospero.model.ProsperoConfig;
+import org.wildfly.prospero.test.MetadataTestUtils;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.wildfly.prospero.api.InstallationMetadata.MANIFEST_FILE_NAME;
+import static org.wildfly.prospero.api.InstallationMetadata.METADATA_DIR;
+
+public class CloneTest extends WfCoreTestBase {
+
+    private Path targetDir;
+    private Path exportPath;
+    private Path importDir;
+
+    @Before
+    public void setUp() throws Exception {
+        super.setUp();
+        targetDir = temp.newFolder().toPath();
+        exportPath = temp.newFolder("export").toPath().resolve("out.zip");
+        importDir = temp.newFolder("restore").toPath().resolve("to_be_restored");
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        if (Files.exists(targetDir)) {
+            FileUtils.deleteDirectory(targetDir.toFile());
+        }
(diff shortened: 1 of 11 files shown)


integration-tests/src/test/java/org/wildfly/prospero/it/cli/CloneTest.java
```diff
@@ -0,0 +1,170 @@
...
+              .anyMatch(m -> m.getGroupId().equals("org.wildfly.core")
+                && m.getArtifactId().equals("wildfly-cli")
+                && m.getVersion().equals("19.0.0.Beta11")));
+
+            // check prosperoConfig
+            assertEquals(1, prosperoConfig.getChannels().size());
+            Channel channel = prosperoConfig.getChannels().get(0);
+            assertEquals("channel-0", channel.getName());
+            assertEquals(3, channel.getRepositories().size());
+            channel.getRepositories().stream().anyMatch(p -> p.getId().equals("maven-central"));
```
spyrkob, line 131: missing `assertTrue`?

also AssertJ Assertions.assertThat has really nice API for testing collections, that can become something like:
```
 assertThat(channel.getRepositories())
  .map(Repository::getId())
  .containsExactly("maven-central", "nexus", maven-redhat-ga");
 ```

File: integration-tests/src/test/java/org/wildfly/prospero/it/cli/CloneTest.java [hunk 2]
```diff
@@ -0,0 +1,170 @@
...
+            CliConstants.ARG_PATH, exportPath.toString()
+          )
+          .execute()
+          .assertReturnCode(ReturnCodes.SUCCESS);
+        checkMetaData(new InstallationMetadata(importDir));
+    }
+
+    @Test
+    public void testImportOverrideRepositories() throws Exception {
+        mockInstallationMeta(targetDir);
```
spyrkob, line 158: Could we modify the `${targetDir}/.installation/manifest.yaml`  so that the `org.wildfly.core:wildfly-cli` stream has version `19.0.0.Beta12-SNAPSHOT`? Then we can check that the temporary repository is indeed used?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/CloneCommand.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @CommandLine.Command(name = CliConstants.Commands.EXPORT)
+    private static class CloneExportCommand extends AbstractCommand {
+
+        @CommandLine.Option(names = CliConstants.DIR, order = 1)
+        Optional<Path> directory;
+
+        @CommandLine.Option(names = CliConstants.ARG_PATH, required = true, paramLabel = CliConstants.PATH, order = 2)
+        String outPath;
```
spyrkob, line 67: Is there a reason that's a String rather then a Path?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/CloneCommand.java
```diff
@@ -0,0 +1,125 @@
...
+
+        CloneExportCommand(Console console, ActionFactory actionFactory) {
+            super(console, actionFactory);
+        }
+
+        @Override
+        public Integer call() throws Exception {
+            Path installationDirectory = determineInstallationDirectory(directory);
+            InstallationExportAction exportAction = new InstallationExportAction(installationDirectory);
```
spyrkob, line 76: Could you move the creation of both `InstallationExportAction` and `InstallationRestoreAction` to ActionFactory? It makes it easier to unit test the commands

prospero-cli/src/main/resources/UsageMessages.properties
```diff
@@ -51,6 +51,10 @@ prospero.channel.remove.usage.header = Unsubscribe the installation from a chann
 prospero.channel.initialize.usage.header = Add a custom channel to be used by the server
 prospero.channel.promote.usage.header = Promote a bundle of artifacts to a custom repository
 
+prospero.clone.usage.header = Manage the installation metadata.
```
spyrkob, line 54: Maybe something like:
```
prospero.clone.usage.header=Operations to export the installation metadata and use it to recreate installation
prospero.clone.export.usage.header=Export all metadata needed to recreate installation
prospero.clone.import.usage.header=Recreate a server based on exported metadata
```

Actually thinking about it maybe the `import` operation should be called `recreate` or `restore`? WDYT?

prospero-common/src/main/java/org/wildfly/prospero/actions/InstallationRestoreAction.java
```diff
@@ -49,13 +53,23 @@ public InstallationRestoreAction(Path installDir, MavenSessionManager mavenSessi
     }
 
     public void restore(Path metadataBundleZip)
```
spyrkob, line 55: can we remove this method and leave only the one with remoteRepositories?

prospero-common/src/main/java/org/wildfly/prospero/actions/InstallationRestoreAction.java
```diff
@@ -49,13 +53,23 @@ public InstallationRestoreAction(Path installDir, MavenSessionManager mavenSessi
...
+
+    public void restore(Path metadataBundleZip, List<Repository> remoteRepositories)
             throws ProvisioningException, IOException, OperationException {
         if (installDir.toFile().exists()) {
             throw Messages.MESSAGES.installationDirAlreadyExists(installDir);
         }
 
         try (final InstallationMetadata metadataBundle = InstallationMetadata.importMetadata(metadataBundleZip)) {
             final ProsperoConfig prosperoConfig = metadataBundle.getProsperoConfig();
+            List<Channel> channels = new ArrayList<>(prosperoConfig.getChannels());
```
spyrkob, line 68: nitpick: s/channels/originalChannels/

spyrkob (changes requested):
Looks good @gaol, just a few nitpicks :)

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/CloneCommand.java
gaol: Because `InstallationExportAction.export(String exportName)` is using `String`, I think we can change the export to use `Path` as the parameter, WDYT?

prospero-cli/src/main/resources/UsageMessages.properties
gaol: I prefer `restore` which sounds more like creating a server from an existing metadata.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/CloneCommand.java
spyrkob: Ah, yes please, let's change `InstallationExportAction.export` 👍

spyrkob (approved):

spyrkob:
Thanks @gaol !

Issue #205, opened 2022-11-30
Provide bash autocompletion


TomasHofman:
The picocli has an ability to generate bash autocompletion script.

This could be added as a separate command: `prospero completion <shell>` (similarly like `oc completion <shell>` or `tkn completion <shell>` work).

PR #207, opened 2022-12-01
Completion command to generate bash/zsh completion script

TomasHofman:
Resolves https://github.com/wildfly-extras/prospero/issues/205


TomasHofman:
CC @spyrkob 

Couple of points to this:

1. The completion of predefined FP names works :+1: .
2. Probably we need to update the main command name, currently it's set to `prospero` (i.e. this is the command name that trigger the completion in bash). What is the final script name that prospero will use? I remember something like `eap-installation-manager.sh`? Are going to update this later on, or shall we do it now?
3. The completion in bash only triggers for predefined script names. E.g. with current settings, `prospero <TAB><TAB>` provides completion, but `./prospero <TAB><TAB>` does not. This may be somewhat problematic, because I assume prospero would typically be called by relative path, not as "global" command.
  The relative path like `bin/prospero` can be configured too, but it cannot contain `./`, as such substrings are sanitized away. If we wanted to support `./prospero` or `./bin/prospero` to trigger completion, we will probably need to generate the completion script a little differently, as it's not supported out of the box.

TomasHofman:
Examples of help messages:

```
$ ./prospero -h
Usage: prospero [-hv] [COMMAND]

Options:
  -h, --help      Display this help message.
  -v, --version   Print prospero version and exit.

Commands:
  install     Install a new application server instance.
  update      Apply the latest available patches on a server instance.
  history     List previous installation states.
  revert      Reverts to a previous installation state.
  channel     Manage channels the installation is subscribed to.
  completion  Generate bash completion script. (Apply with `source <(prospero completion)`)
  clone       Manage the installation metadata.

Exit codes:
  0   Successful program execution.
  1   Failed operation.
  2   Invalid arguments were given.

Use `prospero <COMMAND> --help` to show usage information about given command.
```
```
$ ./prospero completion -h
Generate bash completion script. (Apply with `source <(prospero completion)`)

Usage: prospero completion [-hV]
Generate bash/zsh completion script for prospero.
Run the following command to give `prospero` TAB completion in the current shell:

   source <(prospero completion)

Alternatively, you can save the completion script to a file, and evaluate this file
from your .bashrc script:

   prospero completion > ~/prospero_completion
   echo ". ~/prospero_completion" >> ~/.bashrc


Options:
  -h, --help      Display this help message.
  -V, --version   Print prospero version and exit.
```

prospero-cli/src/main/resources/UsageMessages.properties
```diff
@@ -55,6 +55,19 @@ prospero.clone.usage.header = Manage the installation metadata.
...
+prospero.completion.usage.header = Generate bash completion script. (Apply with `source <(${prospero.dist.name} completion)`)
+prospero.completion.usage.description.0 = Generate bash/zsh completion script for ${ROOT-COMMAND-NAME:-the root command of this command}.
+prospero.completion.usage.description.1 = Run the following command to give `${ROOT-COMMAND-NAME:-$PARENTCOMMAND}` TAB completion in the current shell:
+prospero.completion.usage.description.2 =
+prospero.completion.usage.description.3 = \u0020  source <(${PARENT-COMMAND-FULL-NAME:-$PARENTCOMMAND} ${COMMAND-NAME})
+prospero.completion.usage.description.4 =
+prospero.completion.usage.description.5 = Alternatively, you can save the completion script to a file, and evaluate this file
+prospero.completion.usage.description.6 = from your .bashrc script:
+prospero.completion.usage.description.7 =
+prospero.completion.usage.description.8 = \u0020  ${PARENT-COMMAND-FULL-NAME:-$PARENTCOMMAND} ${COMMAND-NAME} > ~/prospero_completion
```
spyrkob, line 67: s/prospero_completion/${prospero.dist.name}_completion/

TomasHofman: :+1:

TomasHofman:
In fact I don't know if the `Commands.MAIN_COMMAND` is really used for anything that is visible. It could be in theory removed (maybe).

TomasHofman:
Rebased and updated to respect the `prospero.dist.name` property. I think it's now ready.

PR #208, opened 2022-12-01
Improve invalid installation path message: suggest --dir option

TomasHofman:
... if path is set to current directory


prospero-cli/src/main/java/org/wildfly/prospero/cli/CliMessages.java
```diff
@@ -120,6 +121,10 @@ public interface CliMessages {
     @Message("Path `%s` does not contain a server installation provisioned by prospero.")
     IllegalArgumentException invalidInstallationDir(Path path);
 
+    @Message("Path `%s` does not contain a server installation provisioned by prospero."
```
spyrkob, line 124: Let's not refer to prospero in error messages as the name will be changed in the product

TomasHofman: I added another commit that makes this parametrized and obtains the correct dist name from a resource bundle.

PR #210, opened 2022-12-07
Remove references to community-universe

spyrkob:


TomasHofman:
Does this mean if the universe is not mentioned in the FPL it doesn't need to be in the channel?

spyrkob:
Yes, exactly. It's related to GAL-341

PR #213, opened 2022-12-13
Fix incorrect manifest url in channel example.

soul2zimate:
Fix incorrect manifest url in channel example.


soul2zimate:
Applied this change locally, It's surprise to me that running `./prospero install --fpl org.wildfly:wildfly-galleon-pack --channel examples/wildfly-27.0.0.Alpha2-channel.yaml --dir installation-dir` successfully provision a server. The `wildfly-27.0.0.Alpha2-manifest.yaml` contains vendor information which shouldn't be part of manifest file, I actually expect a parsing error during schema validation.

TomasHofman:
Hi @soul2zimate, can you remove the vendor block from the manifest too then?

spyrkob:
Thanks @soul2zimate

Issue #88, opened 2022-07-01
Warn user if any conflicts detected


spyrkob:
Probably will need to modify Galleon for this.

Example: if user modified file 'standalone.xml' and the update modified 'standalone.xml', before the old server is replaced with updated one, we should prompt user whether he's happy to continue.

spyrkob:
Galleon issue: https://issues.redhat.com/browse/GAL-338

jfdenise:
With the redesign of the update (WIP branch: https://github.com/jfdenise/prospero/tree/future), such logic should be implemented directly in Prospero.

PR #214, opened 2022-12-16
Update improvements - split update prepare/apply and artifact cache

### LINKED ISSUE #212: Integrate prepare-update/apply-update
The update operation is split into prepare-update step which provisions a new server and apply-update which merges the changes into the server.

Additionally implement an artifact cache using provisioned artifacts from the server.

spyrkob:
Fixes #212 

**Note: requires changing galleon and galleon-plugins versions before merging**


prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
```diff
@@ -129,14 +151,32 @@ private void performUpdate(UpdateAction updateAction) throws ArtifactResolutionE
...
+        }
+
+        if (!yes && !console.confirmBuildUpdates()) {
+            return;
+        }
+
+        updateAction.buildUpdate(updateDirectory.get().toAbsolutePath());
+
+        console.buildUpdatesComplete();
+    }
```
TomasHofman, line 176: Could we print conflicting files during `buildUpdate()` too? For user convenience...

prospero-cli/src/main/resources/UsageMessages.properties
```diff
@@ -74,6 +76,8 @@ prospero.completion.usage.description.8 = \u0020  ${PARENT-COMMAND-FULL-NAME:-$P
 prospero.completion.usage.description.9 = \u0020  echo \". ~/${prospero.dist.name}_completion\" >> ~/.bashrc
 prospero.completion.usage.description.10 =
 
+prospero.apply-update.usage.header = Apply previously prepared update to the server. The server has to be stopped during this operation.
```
TomasHofman, line 79: Thinking out loud - are we able to detect server is running? Are we doing it?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
spyrkob: We could, but I'm not sure if it belongs at buildUpdate stage. There might be further changes before user does applyUpdate - would we want to display them twice?

prospero-cli/src/main/resources/UsageMessages.properties
spyrkob: Yes, it's the same mechanic Galleon is using: https://github.com/wildfly-extras/prospero/blob/1faa119ab5c220299fe4d5d4d0019580ffe35ba2/prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyUpdateAction.java#L160

prospero-common/src/test/java/org/wildfly/prospero/utils/filestate/PathState.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.IOException;
+import java.io.StringWriter;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/*
+ * Taken from https://github.com/wildfly/galleon/tree/1fa07afb29b2ec61222e746b107a6822e836d05c/core/src/test/java/org/jboss/galleon/test/util/fs/state
```
TomasHofman, line 30: Unrelated to this PR, I was just thinking about how to correctly credit other people's code. I think include original authors name (e.g. via `@author` tag) is a good practice. Link to original code is good. Up to your consideration...

TomasHofman (approved):

TomasHofman:
When looking at current main usage screen:

```
$ ./prospero
...
Commands:
  ...
  update        Apply the latest available patches on a server instance.
  apply-update  Apply previously prepared update to the server. The server has to be stopped during this operation.
  ...
```

I'm not sure if it communicates the mechanics of building & applying updates well. Maybe building an update would deserve it's own command.

TomasHofman:
BTW I'm getting "ERROR: The following plugin options are not recognized: jboss-resolved-artifacts-cache" when running install. I tried to upgrade `version.org.wildfly.galleon-plugins` to 6.2.2.Final but that doesn't help either.

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/ApplyUpdateCommandTest.java
```diff
@@ -0,0 +1,173 @@
...
+    public void targetFolderNeedsToBeProsperoInstallation() throws Exception {
+        final Path updatePath = mockInstallation("update");
+        final Path targetPath = mockInstallation("target");
+        int exitCode = commandLine.execute(CliConstants.Commands.APPLY_UPDATE,
+                CliConstants.UPDATE_DIR, updatePath.toString(),
+                CliConstants.DIR, targetPath.toString());
+
+        Assert.assertEquals(getErrorOutput(), ReturnCodes.SUCCESS, exitCode);
+        verify(applyUpdateAction).applyUpdate();
+    }
```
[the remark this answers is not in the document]
spyrkob: good catch, thanks

prospero-common/src/main/java/org/wildfly/prospero/galleon/GalleonUtils.java
```diff
@@ -47,12 +49,14 @@ public class GalleonUtils {
     public static final String JBOSS_FORK_EMBEDDED_PROPERTY = "jboss-fork-embedded";
     public static final String JBOSS_FORK_EMBEDDED_VALUE = "true";
     public static final String JBOSS_BULK_RESOLVE_PROPERTY = "jboss-bulk-resolve-artifacts";
-    public static final String JBOSS_BULK_RESOLVE_VALUE = "true";
+    public static final String JBOSS_BULK_RESOLVE_VALUE = "false";
```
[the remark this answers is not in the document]
spyrkob: ah, good catch, that should not be changed. Initially CachedVersionResolver only worked on single artifact, I implemented buld operation later and forgot to switch this back

spyrkob:
> I'm not sure if it communicates the mechanics of building & applying updates well. Maybe building an update would deserve it's own command.

Yes I'm not sure what's the best way is tbh. The revert command will have to be split in similar way, so we'd need to add a lot of top level commands

> BTW I'm getting "ERROR: The following plugin options are not recognized: jboss-resolved-artifacts-cache" when running install. I tried to upgrade version.org.wildfly.galleon-plugins to 6.2.2.Final but that doesn't help either.

Are you using the channel files from examples folder? They're using manifests hosted in the repository rather then the updated ones so would use old version of the plugin by default

TomasHofman:
> Are you using the channel files from examples folder?

Yes, that was it :).

>> Could we print conflicting files during buildUpdate() too? For user convenience...

> would we want to display them twice?

Maybe some of the information could be printed in both commands. I also think about the list of updated artifacts. And maybe prompt during `apply-update` whether to really proceed? It would provide a more informative experience :).

> How about update subcommands like:
> ```
> update perform  (in-place?)
> update prepare
> update apply
> ```

The sub-commands look interesting too.

We could leave these things for subsequent PRs.

spyrkob:
> And maybe prompt during apply-update whether to really proceed? It would provide a more informative experience :).

Yes I thought that can be done together with https://github.com/wildfly-extras/prospero/issues/88

spyrkob:
UX improvements in follow up issues: #88 and #217

PR #224, opened 2023-01-13
Clear properties to fix forked process runner in WFLY env

spyrkob:
When running prospero from within WFLY (as CLI operation) property `javax.management.builder.initial` causes ClassNotFoundException


spyrkob:
FYI @yersan

PR #228, opened 2023-01-20
Force to reset system properties for embedded servers by default

yersan:
It requires WFGP-239 via a galleon plugin update:

https://github.com/wildfly/galleon-plugins/pull/243


yersan:
I just did a rebase, no changes.

spyrkob:
Fixed in #264

PR #246, opened 2023-02-03
Generate apply-update command via SPI and upgrade installation Manager API to 1.0.0.Beta2

yersan:
Built on top of #244


spyrkob:
Thanks @yersan !

PR #252, opened 2023-02-08
Fix README running instructions

xstefank:


examples/wildfly-27.0.0.Alpha2-manifest.yaml
```diff
@@ -414,7 +414,7 @@ streams:
 
   - groupId: "io.undertow"
     artifactId: "undertow-core"
-    version: "2.2.17.Final"
+    version: "2.2.18.Final"
```
spyrkob, line 417: @xstefank that shouldn't be committed right?

xstefank: correct, thanks

spyrkob:
thanks @xstefank !

Issue #254, opened 2023-02-09
Change returned codes to represent changes made


spyrkob:
Errors:
2+ - errors

Return different status code in update prepare and rollback operations when no update available
0 - no update
1 - updates found

Return different status code in install operations
0 - installation already exists
1 - created installation

PR #259, opened 2023-02-13
[#254] Change return codes to indicate operations without changes

spyrkob:


spyrkob:
@rpelisse I think that should fix the return codes for ansible. Can you take a look if that makes sense https://github.com/wildfly-extras/prospero/pull/259/files#diff-b6aec5c445003bbfea1760c2b781a0421a0040cf3df585c5ceb0fc6239af1a7b ?

PR #265, opened 2023-02-15
Exclude locally installed artifacts from deployment

spyrkob:


spyrkob:
superseded by https://github.com/wildfly-extras/wildfly-channel/pull/146

PR #288, opened 2023-02-22
[#287] Fix NPE when using MavenOptions.Builder with a null localCache…

### LINKED ISSUE #287: [SPI] NPE when using a MavenOptions with a null local repository
At this moment, the way to specify we do not want to use a local maven repository is to create maven options with a null Path for the local repository.

This is now throwing an NPE when using the Builder at:
https://github.com/wildfly-extras/prospero/blob/main/prospero-common/src/main/java/org/wildfly/prospero/api/MavenOptions.java#L171

yersan:
…Path

Fixes #287


yersan:
Close, this is an invalid fix, @spyrkob will fix this correctly

PR #328, opened 2023-03-17
Update examples.done message

rdnovell:
When prospero run I can see in the log:
```
Installing JBoss examples
Installing JBoss examples APPLYING CHANGES...
Installing JBoss examplesInstalling config standalone-xts.xml
Downloaded artifacts.
```
I think is better to show at the end "JBoss examples installed"


spyrkob:
@rdnovell thanks yes, that's fixed in #324

PR #330, opened 2023-03-20
Corrections to help screen text

spyrkob:
Additionally a few changes to the operations/attributes:
`clone restore` -> `clone recreate`
`print-licences` -> `print-licenses`
`--accept-agreements` -> `--accept-license-agreements`


spyrkob:
@honza-kasik FYI a few changes to the operations/attribute names suggested in the review (see in [prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/CliConstants.java](https://github.com/wildfly-extras/prospero/pull/330/files#diff-1a4b5e92e06d5df0655c5fcd4c0b19f42d0959ade1d10b66b6e8924e1a79aa4f))

spyrkob:
@yersan I don't think of this changes affect the Wildfly integration, but could you take a quick look?

yersan:
Hi @spyrkob, the integration doesn't use the help pages or output generated by Prospero, so we should be fine with this.

yersan:
Yes, right @spyrkob , I confirm we are not using those commands

Issue #349, opened 2023-04-04
The current `prospero-known-combinations.yaml` can not install WildFly 26 correctly.


liweinan:
The current repositories in the yaml file doesn't contain the Red Hat repo so it can't install WildFly 26 correctly:

```bash
⋊> ~/w/prospero on main ⨯ ./prospero install --profile=wildfly --dir=wfly-26 --manifest=examples/wildfly-26.0.0.Final-manifest.yaml                                         00:51:07
+ '[' -n '' ']'
+ JAVA_OPTS=' -Dhttp.proxyHost=localhost -Dhttp.proxyPort=8829 -Dhttps.proxyHost=localhost -Dhttps.proxyPort=8829'
+ '[' -z '' ']'
++ realpath ./prospero
+ full_path=/Users/weli/works/prospero/prospero
++ dirname /Users/weli/works/prospero/prospero
+ PROSPERO_HOME=/Users/weli/works/prospero
+ CLASSPATH=("${PROSPERO_HOME}/prospero-cli/target/"prospero-cli-*-shaded.jar)
+ java -Dhttp.proxyHost=localhost -Dhttp.proxyPort=8829 -Dhttps.proxyHost=localhost -Dhttps.proxyPort=8829 -jar /Users/weli/works/prospero/prospero-cli/target/prospero-cli-1.0.0.Beta9-SNAPSHOT-shaded.jar install --profile=wildfly --dir=wfly-26 --manifest=examples/wildfly-26.0.0.Final-manifest.yaml
Installing profile: wildfly
Using channels:
# manifest: file:/Users/weli/works/prospero/examples/wildfly-26.0.0.Final-manifest.yaml
  repositories:
    id: central
    url: https://repo1.maven.org/maven2/
    id: jboss-public
    url: https://repository.jboss.org/nexus/content/groups/public/

Feature-packs resolved.                                                                           
Packages installed.               
Downloading artifacts

ERROR: Unable to resolve artifacts:
  * org.picketbox:picketbox:jar:5.0.3.Final-redhat-00008 [missing]
  * org.picketbox:picketbox-infinispan:jar:5.0.3.Final-redhat-00008 [missing]
 Attempted repositories:
  *central::https://repo1.maven.org/maven2/
  *jboss-public::https://repository.jboss.org/nexus/content/groups/public/
```

PR #350, opened 2023-04-04
The current prospero-known-combinations.yaml can not install WildFly …

liweinan:
…26 correctly. #349


liweinan:
https://github.com/wildfly-extras/prospero/issues/349

spyrkob (approved):
thanks @liweinan !

Issue #206, opened 2022-11-30
Update the `channel add` command to accept channel as well as manifest


TomasHofman:
The `install` command is accepting manifest or channels. The `channel add` command should offer the same options.

TomasHofman:
Hello @parsharma, I think this issue is still relevant.

With the `install` command we have these two options:

```
# reference a channel:
./prospero install --fpl org.wildfly.core:wildfly-core-galleon-pack::zip --dir wfcore --channels examples/wildfly-core-channel.yaml

# or reference a manifest and repos:
./prospero install --fpl org.wildfly.core:wildfly-core-galleon-pack::zip --dir wfcore2 --manifest examples/wildfly-core-manifest.yaml --repositories central::https://repo1.maven.org/maven2/
```

With the `channel add` command, we can only reference manifest and repositories:

```
./prospero channel add --channel-name channel-1 --manifest examples/wildfly-core-manifest.yaml --repositories central::https://repo1.maven.org/maven2/ --dir wfcore2
```

but we could also have the option to reference just channel, just like in the `install` command.

Do you have any objections @spyrkob ?

spyrkob:
Sounds good. A couple comments:

- the `--channel` should be mutually exclusive with `--manifest` and `--repositories`
- the YAML file can contain multiple channels - I suppose to be consistent with `install` all should be added
- all the added channels have to have a name present. That has to be validated before any of the channels are added

PR #391, opened 2023-05-18
Update the channel add command to accept channel as well as manifest …

parsharma:
Fixes: https://github.com/wildfly-extras/prospero/issues/206

@TomasHofman @spyrkob 
I have sent this PR for review.  Please let me know your views. Work is still going on.

Thanks


prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -58,16 +69,42 @@ public ChannelAddCommand(CliConsole console, ActionFactory actionFactory) {
...
 
-        console.println(CliMessages.MESSAGES.subscribeChannel(installationDirectory, channelName));
+            console.println(CliMessages.MESSAGES.subscribeChannel(installationDirectory, channelName));
 
-        ChannelManifestCoordinate manifest = ArtifactUtils.manifestCoordFromString(gavUrlOrPath);
-        Channel channel = new Channel(channelName, null, null, repositories, manifest, null, null);
+            ChannelManifestCoordinate manifest = ArtifactUtils.manifestCoordFromString(gavUrlOrPath);
+            channels = List.of(new Channel(channelName, null, null, repositories, manifest, null, null));
+        } else {
+            return ReturnCodes.INVALID_ARGUMENTS;
```
TomasHofman, line 89: We should probably print some error message (`console.error(...)`) explaining that either manifest location or channel location must be set. We should also reject both of these parameters being set at once, they should be exclusive.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -58,16 +69,42 @@ public ChannelAddCommand(CliConsole console, ActionFactory actionFactory) {
...
         Path installationDirectory = determineInstallationDirectory(directory);
         List<Repository> repositories;
-        repositories = RepositoryDefinition.from(repositoryDefs);
+        List<Channel> channels = null;
+            if(channelUrlOrPath.isPresent()) {
+                final ChannelCoordinate coord = ArtifactUtils.channelCoordFromString(channelUrlOrPath.get());
+                checkFileExists(coord.getUrl(), channelUrlOrPath.get());
+                ProvisioningDefinition provisioningDefinition = ProvisioningDefinition.builder()
+                        .setChannelCoordinates(channelUrlOrPath.get())
+                        .build();
```
TomasHofman, line 78: We could maybe add the`.setOverrideRepositories(RepositoryDefinition.from(repositoryDefs))` call to the builder, which would allow user to override repositories set in a YAML file.

TomasHofman:
@spyrkob we were thinking yesterday how to handle multiple channels (whether there are multiple documents in provided YAML or even multiple channel locations are given).

In my mind the problem is in channel names. The channel name parameter in the YAML spec is optional, so the input can contain one or more channels without names, in which case we must allow user to set the names or auto-generate them.

We have `--channel-name` attribute, but it's weird to combine it with multiple channels scenario. User may not even be aware how many channels there are in the file...

I was thinking the options are:
* Limit the command to only add the first channel from the YAML, print warning if there are more saying only the first is being added. Setting / overriding the name with `--channel-name` is easy here.
* Allow multiple channels being added, and don't allow `--channel-name` in combination with `--channel` - always use the names from the YAML file, or auto-generate them if missing (and print generated names).

I think it should be kept somehow simple...

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -20,36 +20,47 @@
...
 @CommandLine.Command(name = CliConstants.Commands.ADD, sortOptions = false)
 public class ChannelAddCommand extends AbstractCommand {
 
-    @CommandLine.Option(names = CliConstants.CHANNEL_NAME, required = true)
+    @CommandLine.Option(names = CliConstants.CHANNEL_NAME)
     private String channelName;
 
-    @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST, required = true)
+    @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST)
     private String gavUrlOrPath;
```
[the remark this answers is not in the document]
parsharma: Sure, will change the names.

spyrkob:
@TomasHofman I think the 2nd option sounds good. If the channels in the file do not have names, generate them and prompt the user to accept. Agree that `--channel-name` should be ignored/not allowed when combined with the YAML file

parsharma:
@TomasHofman @spyrkob I apologize for the delay. I was occupied with my responsibilities as an RC and was unable to continue working on this PR. I am now resuming my work.
Thanks

spyrkob:
@parsharma your latest push seems to have removed your changes to ChannelAddCommand class. Was that intentional?

parsharma:
@spyrkob Yes, I need to incorporate my changes from the beginning since the entire file has been modified. However, I couldn't work on it due to some other issues and I am also going on leave.
@moulali intends to take on this task, so he will begin working on it.
Sorry for the delay on this issue.

parsharma:
I have pushed my changes, so @moulalis can use it as a reference.

spyrkob:
@parsharma do you want to continue working on this issue?

parsharma:
@spyrkob yes, sure. Just give me sometime I am still setting up my laptop.

parsharma:
Closing this PR because a new [branch](https://github.com/wildfly-extras/prospero/pull/668) has been created for the same issue. 
Thanks

PR #442, opened 2023-08-17
Add licenses to the prospero feature packs

spyrkob:
Add license information into docs/licenses folder


dist/common/src/main/resources/packages/docs.licenses/pm/wildfly/tasks.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" ?>
+
+<tasks xmlns="urn:wildfly:wildfly-feature-pack-tasks:3.1">
+    <transform stylesheet="docs/licenses/prospero-licenses.xsl" src="docs/licenses/${prospero.dist.name}-feature-pack-licenses.xml" output="docs/licenses/${prospero.dist.name}-feature-pack-licenses.html" feature-pack-properties="true" phase="FINALIZING"/>
+    <delete path="docs/licenses/prospero-licenses.xsl"/>
```
spyrkob, line 5: @jfdenise in the wildfly's version the renamed licenses.xsl is removed - do you know why that's done and if Prospero should also remove the XSL file?

jfdenise: I suspect it is removed because useless inside an installation. Prospero should also remove it I think.

Issue #411, opened 2023-06-06
When displaying the update candidates, display the name of the channel the update was found in


spyrkob:
The example output would be:

```
Updates found:
  org.yaml:snakeyaml                                    1.29                 ==>  1.33  [EAP 8.0]
```

Note: `ChannelSession` in wildfly-channel will have to be modified to expose this information.
Note2: The update candidate would have to preserve this information to display diff correctly.

moulalis:
Hi @spyrkob could you please assign this to me?

PR #447, opened 2023-08-29
[issue:411] When displaying the update candidates, display the name of the channel the update was found in

moulalis:
Fixes: https://github.com/wildfly-extras/prospero/issues/411

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
@@ -17,17 +17,6 @@
 
 package org.wildfly.prospero.cli.commands;
 
-import java.io.IOException;
-import java.net.MalformedURLException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.StandardCopyOption;
-import java.util.List;
-import java.util.Optional;
-import java.util.function.Supplier;
-import java.util.stream.Collectors;
-
 import org.apache.commons.io.FileUtils;
 import org.jboss.galleon.ProvisioningException;
 import org.jboss.galleon.config.ProvisioningConfig;
@@ -65,6 +54,16 @@ import org.wildfly.prospero.updates.UpdateSet;
 import picocli.CommandLine;
 
 import javax.xml.stream.XMLStreamException;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
 
 @CommandLine.Command(
         name = CliConstants.Commands.UPDATE,
@@ -189,7 +188,7 @@ public class UpdateCommand extends AbstractParentCommand {
 
             try (UpdateAction updateAction = actionFactory.update(installationDir,
                     mavenOptions, console, repositories)) {
-                if (buildUpdate(updateAction, candidateDirectory, yes, console, ()->console.confirmBuildUpdates())) {
+                if (buildUpdate(updateAction, candidateDirectory, yes, console, () -> console.confirmBuildUpdates(), installationDir)) {
                     console.println("");
                     console.buildUpdatesComplete();
                     console.println(CliMessages.MESSAGES.updateCandidateGenerated(candidateDirectory));
@@ -255,7 +254,7 @@ public class UpdateCommand extends AbstractParentCommand {
             applyCandidateAction.applyUpdate(ApplyCandidateAction.Type.UPDATE);
             console.updatesComplete();
 
-            if(remove) {
+            if (remove) {
                 applyCandidateAction.removeCandidate(candidateDir.toFile());
             }
             final float totalTime = (System.currentTimeMillis() - startTime) / 1000f;
@@ -272,6 +271,7 @@ public class UpdateCommand extends AbstractParentCommand {
         public ListCommand(CliConsole console, ActionFactory actionFactory) {
             super(console, actionFactory);
         }
+
         @Override
         public Integer call() throws Exception {
             final long startTime = System.currentTimeMillis();
@@ -313,7 +313,7 @@ public class UpdateCommand extends AbstractParentCommand {
         public Integer call() throws Exception {
             final Path installDir = directory.orElse(currentDir()).toAbsolutePath();
             if (Files.exists(ProsperoMetadataUtils.manifestPath(installDir))
-              || Files.exists(ProsperoMetadataUtils.configurationPath(installDir))) {
+                    || Files.exists(ProsperoMetadataUtils.configurationPath(installDir))) {
                 console.println(CliMessages.MESSAGES.metadataExistsAlready(installDir, DistributionInfo.DIST_NAME));
                 return ReturnCodes.INVALID_ARGUMENTS;
             }
@@ -340,8 +340,8 @@ public class UpdateCommand extends AbstractParentCommand {
         private void generateMeta(Path installDir, SubscribeNewServerAction.GenerateResult generateResult) throws IOException, ProvisioningException {
             // compare hashes
             FsEntryFactory fsEntryFactory = FsEntryFactory.getInstance()
-              .filterGalleonPaths()
-              .filter(ProsperoMetadataUtils.METADATA_DIR);
+                    .filterGalleonPaths()
+                    .filter(ProsperoMetadataUtils.METADATA_DIR);
             final FsEntry originalState = fsEntryFactory.forPath(generateResult.getProvisionDir());
             final FsEntry currentState = fsEntryFactory.forPath(installDir);
             final FsDiff diff = FsDiff.diff(originalState, currentState);
@@ -379,7 +379,7 @@ public class UpdateCommand extends AbstractParentCommand {
                 channels = channels.stream().map(c -> {
                     try {
                         return new Channel(c.getName(), c.getDescription(), c.getVendor(), c.getRepositories(),
-                          new ChannelManifestCoordinate(manifestPathCopy.toUri().toURL()), c.getBlocklistCoordinate(), c.getNoStreamStrategy());
+                                new ChannelManifestCoordinate(manifestPathCopy.toUri().toURL()), c.getBlocklistCoordinate(), c.getNoStreamStrategy());
                     } catch (MalformedURLException e) {
                         throw new RuntimeException(e);
                     }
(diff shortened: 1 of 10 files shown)


spyrkob (changes requested):
The missing part is `update/revert apply` operations. They display the differences as well, it would be good to display the channel name there as well

prospero-common/src/main/java/org/wildfly/prospero/updates/UpdateFinder.java
```diff
@@ -82,19 +82,33 @@ public UpdateSet findUpdates(List<Artifact> artifacts) throws ArtifactResolution
...
             return Optional.empty();
         } else {
-            return Optional.of(ArtifactChange.updated(artifact, latest));
+            ArtifactChange change;
+            if (artifact == null) {
+                change = ArtifactChange.added(latest, channelName.map(val -> "[" + val + "]").orElse(""));
+            } else if (latest == null) {
+                change = ArtifactChange.removed(latest);
+            } else {
+                change = ArtifactChange.updated(artifact, latest,channelName.map(val -> "[" + val + "]").orElse(""));
```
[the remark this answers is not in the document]
moulalis: > Why is the channelName wrapped in "[]"? That feels like a display consideration.

It is shown in the requirenment https://github.com/wildfly-extras/prospero/issues/411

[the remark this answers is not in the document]
spyrkob: Yes. But that's how the information is printed. It should not be present in data passed around.

prospero-common/src/main/java/org/wildfly/prospero/updates/UpdateFinder.java
```diff
@@ -82,10 +83,13 @@ public UpdateSet findUpdates(List<Artifact> artifacts) throws ArtifactResolution
     private Optional<ArtifactChange> findUpdates(Artifact artifact) throws ArtifactResolutionException {
 
         final String latestVersion;
+        Optional<String> channelName;
```
spyrkob, line 86: final

prospero-common/src/main/java/org/wildfly/prospero/updates/UpdateFinder.java
```diff
@@ -94,7 +98,7 @@ private Optional<ArtifactChange> findUpdates(Artifact artifact) throws ArtifactR
         if (latestVersion == null || latest.getVersion().equals(artifact.getVersion())) {
             return Optional.empty();
         } else {
-            return Optional.of(ArtifactChange.updated(artifact, latest));
+            return Optional.of(ArtifactChange.updated(artifact, latest, channelName.orElse("")));
```
spyrkob, line 101: if channelName is empty, use the `ArtifactChange.updated(artifact, latest)`.

prospero-cli/src/test/java/org/wildfly/prospero/cli/CliConsoleTest.java
```diff
@@ -0,0 +1,70 @@
...
+    public void setUp() {
+        cliConsole = new CliConsole();
+        outputStream = new ByteArrayOutputStream();
+        System.setOut(new PrintStream(outputStream));
+    }
+
+    @Test
+    public void testUpdatesFoundWithUpdates_ArtifactChange_update() {
+        List<ArtifactChange> artifactChanges = new ArrayList<>();
+        ArtifactChange artifactChange = ArtifactChange.updated(new DefaultArtifact("test.group", "test-artifact2", "jar", "2.0.0"), new DefaultArtifact("test.group", "test-artifact2", "jar", "2.1.0"), "[channel-1]");
```
spyrkob, line 49: the channel name should not have braces around it

moulalis:
Do not review, I am looking at some Tests failures.

moulalis:
Hi @spyrkob ,
I've updated the PR, please review it.
Thanks.

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyCandidateAction.java
```diff
@@ -125,6 +131,35 @@ public ApplyCandidateAction(Path installationDir, Path updateDir)
...
+
+            return galleonEnv.getChannelSession();
+        }catch (UnresolvedChannelMetadataException | MetadataException e){
+            log.debug(e.getMessage());
+            return GalleonEnvironment.builder(installationDir,
+                    Collections.emptyList(),
+                    new MavenSessionManager(mavenOptions)).build().getChannelSession();
+        }
+    }
+    private String getChannelName(Artifact artifact) {
```
spyrkob, line 152: I'm not sure about if this would work. 
1. The channel name might not be possible to be resolved in offline 
2. Between creating the candidate and applying it new versions of artifacts could be published possibly into different channels then listed here. 

The better solution would be to use UpdateSet generated during the building of the candidate to generate a record of updates in the candidate and use that information in apply candidate step.

moulalis: > I'm not sure about if this would work.
> 
> 1. The channel name might not be possible to be resolved in offline
> 2. Between creating the candidate and applying it new versions of artifacts could be published possibly into different channels then listed here.
> 
> The better solution would be to use UpdateSet generated during the building of the candidate to generate a record of updates in the candidate and use that information in apply candidate step.

Hi @spyrkob ,

I've updated the above changes, please let me know if any review changes are required.
Thanks.

prospero-metadata/src/main/java/org/wildfly/prospero/metadata/ProsperoMetadataUtils.java
```diff
@@ -53,6 +53,8 @@ public class ProsperoMetadataUtils {
 
     private static final String WARNING_MESSAGE = "WARNING: The files in .installation directory should be only edited by the provisioning tool.";
 
+    public static final String CANDIDATE = ".candidate.txt";
```
spyrkob, line 56: Could you move the constant to PrepareCandidateAction. prospero-metadata should contain only information needed by external tools to work with prospero, I don't think candidate information is needed.

Also could you rename this to `candidate_artifact_channels.txt` or something similar?

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyCandidateAction.java
```diff
@@ -706,4 +717,24 @@ private static void glold(Path installationFile, final Path target) throws Provi
...
+        final Path candidateFile = installationDir.resolve(ProsperoMetadataUtils.METADATA_DIR).resolve(ProsperoMetadataUtils.CANDIDATE);
+        Map<String, String> getInfo = new HashMap<String, String>();
+
+        try (BufferedReader br = new BufferedReader(new FileReader(candidateFile.toFile()))) {
+            String line;
+            while ((line = br.readLine()) != null) {
+                String[] parts = line.split("=");
+                if (parts.length == 2) {
+                    String dependencyName = parts[0];
+                    String dependencyVersion = parts[1];
```
spyrkob, line 731: is this really dependencyVersion? I though this maps dependencyName to channelName?

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyCandidateAction.java
```diff
@@ -706,4 +717,24 @@ private static void glold(Path installationFile, final Path target) throws Provi
...
+                    getInfo.put(dependencyName, dependencyVersion);
+                }
+            }
+        } catch (IOException e) {
+            System.err.println("An error occurred: " + e.getMessage());
```
spyrkob, line 736: Could you use Log instead of System.err and provide some information about error - Unable to read the channel names file, etc.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
```diff
@@ -434,6 +437,24 @@ private static boolean buildUpdate(UpdateAction updateAction, Path updateDirecto
...
+        if (updateSet.isEmpty()) {
+            return false;
+        }
+
+        if (!yes && !confirmation.get()) {
+            return false;
+        }
+
+        updateAction.buildUpdate(updateDirectory.toAbsolutePath());
+        writeChannelMapToFile(updateSet, installationDir);
```
spyrkob, line 453: Could you move this to `PrepareCandidateAction`? The updateSet is available in `UpdateAction.buildUpdate` and can be passed to the `PrepareCandidateAction`

spyrkob (changes requested):

prospero-metadata/src/main/java/org/wildfly/prospero/metadata/ProsperoMetadataUtils.java
moulalis: > Could you move the constant to PrepareCandidateAction. prospero-metadata should contain only information needed by external tools to work with prospero, I don't think candidate information is needed.
> 
> Also could you rename this to `candidate_artifact_channels.txt` or something similar?

Hi @spyrkob,
This constant is also used in ApplyCandidateAction so it has to be kept in common place.

moulalis:
Hi @spyrkob ,
I've updated the PR, please review it.
Thanks.

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyCandidateAction.java
```diff
@@ -278,12 +281,20 @@ public UpdateSet findUpdates() throws OperationException {
...
         List<ArtifactChange> changes = new ArrayList<>();
+
+        Map<String, String> channelMap = getChannelNameMap();
         for (String key : baseMap.keySet()) {
+
             if (candidateMap.containsKey(key)) {
                 if (!baseMap.get(key).getVersion().equals(candidateMap.get(key).getVersion())) {
-                    changes.add(ArtifactChange.updated(baseMap.get(key), candidateMap.get(key)));
+                    if (channelMap.isEmpty()) {
+                        changes.add(ArtifactChange.updated(baseMap.get(key), candidateMap.get(key)));
```
spyrkob, line 294: I think you should add the change without the channel name if the `channelMap` doesn't contain the key as well. The `isEmpty` check is not really needed.

prospero-common/src/main/java/org/wildfly/prospero/actions/UpdateAction.java
```diff
@@ -116,6 +117,8 @@ public boolean buildUpdate(Path targetDir) throws ProvisioningException, Operati
 
             final boolean result = prepareCandidateAction.buildCandidate(targetDir, galleonEnv,
                     ApplyCandidateAction.Type.UPDATE, provisioningConfig);
+
+            prepareCandidateAction.writeChannelMapToFile(updateSet, installDir);
```
spyrkob, line 121: that should be written to the generated candidate (`targetDir`) not the original installation

prospero-common/src/main/java/org/wildfly/prospero/actions/PrepareCandidateAction.java
```diff
@@ -115,4 +120,24 @@ private void writeProsperoMetadata(Path home, ChannelManifest manifest, List<Cha
...
+        try {
+            FileWriter fileWriter = new FileWriter(candidateFile.toFile());
+            BufferedWriter bufferedWriter = new BufferedWriter(fileWriter);
+
+            for (ArtifactChange artifactChange : updateSet.getArtifactUpdates()) {
+                String channelData = artifactChange.getArtifactName() + "=" + (artifactChange.getChannelName().isEmpty() ? "" : artifactChange.getChannelName().get());
+                bufferedWriter.write(channelData);
+                bufferedWriter.newLine(); // Add a newline after each entry
+            }
+            bufferedWriter.close();
```
spyrkob, line 136: can you move the writer.close to a finally block or use try-with-resources?

prospero-metadata/src/main/java/org/wildfly/prospero/metadata/ProsperoMetadataUtils.java
```diff
@@ -53,6 +53,8 @@ public class ProsperoMetadataUtils {
 
     private static final String WARNING_MESSAGE = "WARNING: The files in .installation directory should be only edited by the provisioning tool.";
 
+    public static final String CANDIDATE = "candidate_artifact_channels.txt";
```
spyrkob, line 56: Could you move this to PrepareCandidateAction please? This file is candidate-specific

PR #526, opened 2023-12-20
Change --no-resolve-local-cache to --use-default-maven-cache and disable maven cache by default

spyrkob:
Upstream #525


yersan (approved):
LGTM

PR #549, opened 2024-01-31
If the provisioning configuration is the same as saved copy, the file is not overwritten (fix duplicate code)

soul2zimate:
The Java doc on the method says it should be `If the provisioning configuration is the same as saved copy, the file is not overwritten.`

This is for 1.1.x code before the `prospero-metadata` split

Upstream PR: https://github.com/wildfly-extras/prospero-metadata/pull/11


spyrkob (approved):

spyrkob:
Thanks @soul2zimate!

PR #553, opened 2024-02-02
JBEAP-26480: add --debug option to CLI

michpetrov:
Issue: [JBEAP-26480](https://issues.redhat.com/browse/JBEAP-26480)

Feels like there should be better way to get a hold of the console handler but I haven't found any.


spyrkob (changes requested):
You can use `LogManager.getLogManager()` to get loggers as well.

This doesn't work when using the prospero distribution (`dist/build/target`). Either the default logging configuration needs to include a console handler set to level OFF, or the code needs to add the console handler.

michpetrov:
Updated.

> You can use `LogManager.getLogManager()` to get loggers as well.

Yeah, but my problem was that we could only access loggers this way. E.g. I cannot directly do `getHandler("CONSOLE").setLevel(DEBUG)` or `getLogger("").addHandler("CONSOLE")` because I cannot access handlers by name. The only solution seems to be accessing the whole configuration the way I did.

Btw, is the `-Pdist` build supposed to be creating a jar? I thought that's what I should run before I found out there is no Java code inside.

spyrkob:
> Btw, is the -Pdist build supposed to be creating a jar? I thought that's what I should run before I found out there is no Java code inside.

No -Pdist builds feature packs and provisions the standalone feature pack to create a distribution of prospero.

michpetrov:
Right, what I'm saying is that if I run -Pdist it generates 3 files - zip, tar.gz and jar.

I've added the check, as far as I can tell it will only fail if someone deletes the properties file. You want this ported to 1.1.x too?

spyrkob (approved):

spyrkob:
> Right, what I'm saying is that if I run -Pdist it generates 3 files - zip, tar.gz and jar.

Ah sorry I misunderstood you. Yes the jar is empty, I'm not sure if it's possible to disable building it

> You want this ported to 1.1.x too?
Yes please if you don't mind

PR #569, opened 2024-03-01
[JBEAP-26296] Allow specifying --repositories …

ChristinaDsl:
https://issues.redhat.com/browse/JBEAP-26296


spyrkob (changes requested):
@ChristinaDsl that doesn't work, the local path has to be converted to an URL to be usable:

```
./prospero install --fpl org.wildfly.core:wildfly-core-galleon-pack:20.0.0.Beta3 \
  --manifest examples/wildfly-core-manifest.yaml --repositories test-repo \
  --dir installation-dir
Installing feature pack: org.wildfly.core:wildfly-core-galleon-pack:20.0.0.Beta3
Using channels:
# manifest: file:/Users/spyrkob/workspaces/set/prospero/prospero/examples/wildfly-core-manifest.yaml
  repositories:
    id: temp-repo-0
    url: test-repo



ERROR: The following artifacts could not be resolved: org.wildfly.core:wildfly-core-galleon-pack:zip:20.0.0.Beta3 (absent): Could not transfer artifact org.wildfly.core:wildfly-core-galleon-pack:zip:20.0.0.Beta3 from/to temp-repo-0 (test-repo): Cannot access test-repo with type default using the available connector factories: BasicRepositoryConnectorFactory
```

Also could you add unit tests to `RepositoryDefinitionTest` and an integration test to `org.wildfly.prospero.it.cli.InstallTest` to verify this change?

prospero-cli/src/test/java/org/wildfly/prospero/cli/RepositoryDefinitionTest.java
```diff
@@ -75,5 +92,35 @@ public void throwsErrorIfFormatIsIncorrect() throws Exception {
...
+    }
+
+    @Test
+    public void testCorrectRelativeOrAbsolutePathForFileURL() throws Exception {
+        try {
+            from(List.of("file:../repo"));
+            from(List.of(fileURL));
+            from(List.of("file://"+filePath));
+        } catch (Exception e) {
+            fail("Provided URL's absolute or relative path is not correct");
```
spyrkob, line 123: @ChristinaDsl you don't need to catch the exception. If it fails the test fails. Also you should check that the returned value is correct. The method might run without exception, but the returned value might be wrong

prospero-cli/src/test/java/org/wildfly/prospero/cli/RepositoryDefinitionTest.java
```diff
@@ -75,5 +92,48 @@ public void throwsErrorIfFormatIsIncorrect() throws Exception {
...
+        assertThrows(ArgumentParsingException.class, ()->from(List.of("file:/path/to/repo")));
+
+        assertThrows(ArgumentParsingException.class, ()->from(List.of(filePath)));
+
+        assertThrows(ArgumentParsingException.class, ()->from(List.of("file://../repo")));
+
+    }
+
+    @Test
+    public void testCorrectRelativeOrAbsolutePathForFileURL() throws Exception {
```
spyrkob, line 117: what about a relative path?

spyrkob (changes requested):

integration-tests/src/test/java/org/wildfly/prospero/it/cli/InstallTest.java
```diff
@@ -55,4 +59,26 @@ public void testInstallWithProvisionConfig() throws Exception {
...
+        MetadataTestUtils.prepareChannel(provisionConfig, List.of(manifestPath.toUri().toURL()));
+
+        install(provisionConfig, targetDir.toPath());
+
+        upgradeStreamInManifest(manifestPath, resolvedUpgradeArtifact);
+
+        final URL temporaryRepo = mockTemporaryRepo(true);
+
+        ExecutionUtils.prosperoExecution(CliConstants.Commands.UPDATE, CliConstants.Commands.PERFORM,
+                        CliConstants.REPOSITORIES, temporaryRepo.toString(),
```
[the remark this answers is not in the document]
ChristinaDsl: Thanks for your comment. This test is to verify that a local repo can be found and the update command is indeed working when an argument " --repositories" is passed along with the url of local repo. So, why should be a relative path here?

prospero-cli/src/test/java/org/wildfly/prospero/cli/RepositoryDefinitionTest.java
ChristinaDsl: The relative path is test in the first lines of this method with this url "file:../prospero-common"

spyrkob: That's a URI with relative path. A relative path would be `../prospero-common` on its own

integration-tests/src/test/java/org/wildfly/prospero/it/cli/InstallTest.java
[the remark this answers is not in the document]
spyrkob: There already are tests in `UpdateWithAdditionalRepositoryTest` that verify this use case. The new use case added in this issue should be that a user can pass a local path (e.g. `../test-repository` or `/home/user/test-repository`) as an argument for `--repositories`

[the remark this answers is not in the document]
ChristinaDsl: So check both for absolute and relative paths in the this method?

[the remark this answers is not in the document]
spyrkob: ideally yes, but relative path would be enough

prospero-cli/src/test/java/org/wildfly/prospero/cli/RepositoryDefinitionTest.java
spyrkob: I believe this verifies JBEAP-26295 as well so those are valid tests. Just adding some examples of paths will be enough.

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -49,10 +55,22 @@ public static List<Repository> from(List<String> repos) throws ArgumentParsingEx
...
     private static boolean isValidUrl(String text) {
+        Path path;
+
         try {
-            new URL(text);
+            URL url = new URL(text);
+            if (text.startsWith("file://..")){
+                return false;
+            } else if (text.startsWith("file")){
+                if (Pattern.compile("/[A-Z]:").matcher(text).find()){
```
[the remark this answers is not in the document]
ChristinaDsl: It does not work on relative paths. In windows and file URLs Paths.get() throws error and in Linux relative path cannot be converted initially in URL.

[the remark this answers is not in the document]
spyrkob: OK thank you. Could you add this explanation in a comment in the code please?

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -49,10 +55,22 @@ public static List<Repository> from(List<String> repos) throws ArgumentParsingEx
     }
 
     private static boolean isValidUrl(String text) {
+        Path path;
+
         try {
-            new URL(text);
+            URL url = new URL(text);
+            if (text.startsWith("file://..")){
+                return false;
```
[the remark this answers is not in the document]
ChristinaDsl: Of course

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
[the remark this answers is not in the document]
ChristinaDsl: Of course

integration-tests/src/test/java/org/wildfly/prospero/it/cli/InstallTest.java
```diff
@@ -55,4 +55,28 @@ public void testInstallWithProvisionConfig() throws Exception {
...
+        MetadataTestUtils.copyManifest("manifests/wfcore-base.yaml", manifestPath);
+        MetadataTestUtils.prepareChannel(provisionConfig, List.of(manifestPath.toUri().toURL()));
+
+        install(provisionConfig, targetDir.toPath());
+
+        upgradeStreamInManifest(manifestPath, resolvedUpgradeArtifact);
+
+        final URL temporaryRepo = mockTemporaryRepo(true);
+
+        Path relativePath = Paths.get(System.getProperty("user.dir")).relativize(Path.of(Path.of(temporaryRepo.toURI()).toUri()));
```
spyrkob, line 72: The nested Path.of is not needed

integration-tests/src/test/java/org/wildfly/prospero/it/cli/InstallTest.java
```diff
@@ -55,4 +55,28 @@ public void testInstallWithProvisionConfig() throws Exception {
...
+
+        ExecutionUtils.prosperoExecution(CliConstants.Commands.UPDATE, CliConstants.Commands.PERFORM,
+                        CliConstants.REPOSITORIES, ".."+File.separator+"integration-tests"+File.separator+relativePath,
```
spyrkob, line 75: You can use Path.of(String...) to build the path instead of appending the strings.

Also, `relativePath` is already a relative path, so I don't think adding `../integration-tests" to it brings anything to the test. I'd keep it simple and just use the `relativePath` here

prospero-cli/src/main/java/org/wildfly/prospero/cli/CliMessages.java
```diff
@@ -418,6 +418,10 @@ default ArgumentParsingException invalidRepositoryDefinition(String repoKey) {
         return new ArgumentParsingException(format(bundle.getString("prospero.general.validation.repo_format"), repoKey));
     }
 
+    default ArgumentParsingException invalidFilePath(String repoKey) {
```
spyrkob, line 421: s/repoKey/path

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -17,43 +17,93 @@
...
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 public class RepositoryDefinition {
 
-    public static List<Repository> from(List<String> repos) throws ArgumentParsingException {
+    public static List<Repository> from(List<String> repos) throws ArgumentParsingException, MalformedURLException, URISyntaxException {
```
spyrkob, line 36: Can you catch the new exceptions and wrap them in ArgumentParsingException with an explanation message?

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -17,43 +17,93 @@
...
+        }
+        if (!text.startsWith("file")) {
+            text = "file:" + text;
+        }
+        return new URL(text).toString();
+    }
+
+    private static boolean isValidUrl(String text) throws ArgumentParsingException {
+        if (text.startsWith("file://..")) {
+            Logger.getLogger("'file://../path/to/repo' is invalid as the format of file URL should be " +
```
spyrkob, line 72: Did you mean Logger.warn ?

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -17,43 +17,93 @@
...
+            throw CliMessages.MESSAGES.invalidFilePath(path.toString());
+        }
+    }
+
+    /* Pattern command, exists in order to recognize Windows file paths properly.
+       For Windows systems, Path.of(Paths.get(url.toURI()).toString()) works correctly,
+       whereas for Linux, Paths.get(url.getPath()).toAbsolutePath() is more suitable. */
+    private static Path getPathFromURL(String text) throws MalformedURLException, URISyntaxException, ArgumentParsingException {
+        URL url = new URL(text);
+        if (Pattern.compile("/[A-Z]:").matcher(text).find()) {
```
spyrkob, line 103: That's not going to correct, e.g. in Linux it's possible to create a folder with name `A:`

integration-tests/src/test/java/org/wildfly/prospero/it/cli/InstallTest.java
spyrkob: `target/folder1/folder2` is a relative path. An absolute path starts with root directory (e.g. `/home/jboss/` on linux or `C:\Users\jboss` on Windows). A relative path doesn't have a root folder and is relative to the current folder (e.g. `target/foo` is relative because it depends on where the user is)

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -56,4 +69,31 @@ private static boolean isValidUrl(String text) {
             return false;
         }
     }
+
+    public static URI getAbsoluteFileURI(String text) throws ArgumentParsingException, URISyntaxException {
```
iweiss, line 73: Same as above.

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -56,4 +69,31 @@ private static boolean isValidUrl(String text) {
...
+        String path;
+        path = getPath(text);
```
iweiss, line 75: Why is the assignment on a separate line?

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
```diff
@@ -241,37 +241,45 @@ public void provisionConfigAndChannelSet() throws IOException {
     @Test
     public void provisionConfigAndRemoteRepoSet() throws Exception {
         Path channelsFile = temporaryFolder.newFile().toPath();
+
+        File test = temporaryFolder.newFolder();
+        String testURL = test.toURI().toURL().toString();
```
iweiss, line 246: Why converting to URI, then to URL and then to String?

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
```diff
@@ -241,37 +241,45 @@ public void provisionConfigAndChannelSet() throws IOException {
...
     }
 
     @SuppressWarnings("unchecked")
     @Test
     public void passShadowRepositories() throws Exception {
         Path channelsFile = temporaryFolder.newFile().toPath();
+
+        File test = temporaryFolder.newFolder();
+        String testURL = test.toURI().toURL().toString();
+
```
iweiss, line 269: Same as above

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
```diff
@@ -241,37 +241,45 @@ public void provisionConfigAndChannelSet() throws IOException {
     @Test
     public void provisionConfigAndRemoteRepoSet() throws Exception {
         Path channelsFile = temporaryFolder.newFile().toPath();
+
+        File test = temporaryFolder.newFolder();
```
iweiss, line 245: Variable could have a more descriptive name

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/InstallCommandTest.java
```diff
@@ -241,37 +241,45 @@ public void provisionConfigAndChannelSet() throws IOException {
...
-                .containsOnly("file:/test");
+                .containsOnly(testURL);
     }
 
     @SuppressWarnings("unchecked")
     @Test
     public void passShadowRepositories() throws Exception {
         Path channelsFile = temporaryFolder.newFile().toPath();
+
+        File test = temporaryFolder.newFolder();
```
iweiss, line 267: Variable could have a more descriptive name

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -56,4 +69,31 @@ private static boolean isValidUrl(String text) {
...
+        File file = new File(path);
```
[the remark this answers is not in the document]
ChristinaDsl: If file does not exist, the error message should include the path, so it is used in more than one line.

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
ChristinaDsl: I will fix this

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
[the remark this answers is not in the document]
iweiss: Do you mean, `throw CliMessages.MESSAGES.invalidFilePath(path);`? If so, you are already using `path` instead of `file`.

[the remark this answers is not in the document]
ChristinaDsl: What do you mean?
file is also used in two lines,
if (file.exists()) {
return file.toURI();

it that is your point..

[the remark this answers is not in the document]
iweiss: My point is that you don't need to create a new variable for the single purpose, `path`, to be used as a parameter for `getPath()` and then used to create a `File` object. You can simplify it with `File file = new File(getPath(text));` for example.

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -56,4 +69,31 @@ private static boolean isValidUrl(String text) {
...
+        if (file.exists()) {
+            return file.toURI();
+        } else {
+            throw CliMessages.MESSAGES.invalidFilePath(path);
+        }
+    }
+
+    public static String getPath(String text) throws URISyntaxException {
+        if (text.startsWith("file")) {
```
iweiss, line 85: I think this should be `file:` to make sure the user is inputting a URI and not a file named `file_dir` for example

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -17,32 +17,45 @@
...
 
+    private static final Logger logger = Logger.getLogger(RepositoryDefinition.class.getName());
+
     public static List<Repository> from(List<String> repos) throws ArgumentParsingException {
         ArrayList<Repository> repositories = new ArrayList<>(repos.size());
         for (int i = 0; i < repos.size(); i++) {
-            final String text = repos.get(i);
-            if(text.contains("::")) {
-                final String[] parts = text.split("::");
+            String repoName = repos.get(i);
```
spyrkob, line 40: nitpick - the `repoName` variable name is a bit misleading - the string can either contain repoName::repoLocation or just repoLocation

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -17,32 +17,45 @@
...
                 if (parts.length != 2 || parts[0].isEmpty() || parts[1].isEmpty() || !isValidUrl(parts[1])) {
-                    throw CliMessages.MESSAGES.invalidRepositoryDefinition(text);
+                    throw CliMessages.MESSAGES.invalidRepositoryDefinition(repoName);
                 }
-
                 repositories.add(new Repository(parts[0], parts[1]));
             } else {
-                if (!isValidUrl(text)) {
-                    throw CliMessages.MESSAGES.invalidRepositoryDefinition(text);
+                if (!repoName.startsWith("http") && !repoName.isEmpty()) {
```
spyrkob, line 48: should check if it starts with either `http://` or `https://`.

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
```diff
@@ -56,4 +69,29 @@ private static boolean isValidUrl(String text) {
...
+            throw CliMessages.MESSAGES.invalidFilePath(repoName);
+        }
+    }
+
+    public static String getPath(String repoName) throws URISyntaxException {
+        if (repoName.startsWith("file:")) {
+            URI inputUri = new URI(repoName);
+            if (inputUri.getScheme().equals("file") &&
+                    !inputUri.getSchemeSpecificPart().startsWith("/")) {
+                inputUri = new URI(("file:/" +
```
spyrkob, line 87: IIUC at this point this is a relative URI. In which case Path.of(inputUri.getSchemeSpecificPart()).toAbsolutePath() should be all you need to get the absolute path

spyrkob: Also this already creates the Path from the URI, why does it need to be converted back to URI and then again to Path?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractInstallCommand.java
```diff
@@ -100,7 +102,7 @@ protected MavenOptions getMavenOptions() throws ArgumentParsingException {
         return mavenOptions.build();
     }
 
-    protected ProvisioningDefinition.Builder buildDefinition() throws MetadataException, NoChannelException, ArgumentParsingException {
+    protected ProvisioningDefinition.Builder buildDefinition() throws MetadataException, NoChannelException, ArgumentParsingException, MalformedURLException, URISyntaxException {
```
spyrkob, line 105: Those exceptions should be wrapped in ArgumentParsingException

spyrkob (changes requested):

prospero-cli/src/main/java/org/wildfly/prospero/cli/RepositoryDefinition.java
ChristinaDsl: if it starts with "https", it should be handled differently?

spyrkob: Both http and https should be treated the same - as web URLs. The difference is that `http://foo.bar` is an URL `http/foo/bar` is a path

spyrkob:
@ChristinaDsl please remove the `[installation-manager]` from commit description, and I think it'll be good to go.

spyrkob (approved):

Issue #580, opened 2024-03-05
SPI Apply update/revert generated commands need to be updated to include the --rm option


yersan:
The apply update/revert commands used by the SPI are generated directly in Prospero.

We need to update them to include the --rm option

PR #579, opened 2024-03-05
[JBEAP-24913, #216] Additional options for update/revert apply operations

spyrkob:
[#216] Additional options for update/revert apply operations,

Issue: https://issues.redhat.com/browse/JBEAP-24913
Upstream: #363

PR #583, opened 2024-03-05
[JBEAP-24913, #580] Add the --rm option to the SPI apply update and revert commands

yersan:
Builds on top of #579

Add the --rm option to get SPI generated commands

Upstream: #582
Issue: https://issues.redhat.com/browse/JBEAP-24913

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/spi/CliProviderImpl.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/spi/CliProviderImpl.java
@@ -53,7 +53,8 @@ public class CliProviderImpl implements CliProvider {
         return CliConstants.Commands.UPDATE + " " + CliConstants.Commands.APPLY + " "
                 + CliConstants.DIR + " " + escape(installationPath.toAbsolutePath()) + " "
                 + CliConstants.CANDIDATE_DIR + " " + escape(candidatePath.toAbsolutePath()) + " "
-                + CliConstants.YES;
+                + CliConstants.YES + " "
+                + CliConstants.REMOVE;
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


prospero-cli/src/main/java/org/wildfly/prospero/cli/spi/CliProviderImpl.java
```diff
@@ -53,15 +53,17 @@ public String getApplyUpdateCommand(Path installationPath, Path candidatePath) {
...
     }
 
     @Override
     public String getApplyRevertCommand(Path installationPath, Path candidatePath) {
         return CliConstants.Commands.REVERT + " " + CliConstants.Commands.APPLY + " "
                 + CliConstants.DIR + " " + escape(installationPath.toAbsolutePath()) + " "
                 + CliConstants.CANDIDATE_DIR + " " + escape(candidatePath.toAbsolutePath()) + " "
-                + CliConstants.YES;
+                + CliConstants.YES + " "
+                + CliConstants.REMOVE;
```
spyrkob, line 66: should this be controlled by a parameter?

yersan: I thought about that but ended up it could be the same case as `CliConstants.YES`. In addition, I did not find it useful for now in our current use case. After applying the candidate server, we should delete it.

I don't mind if it gets parametrized, but found it unuseful now, so went with the easiest path.

yersan:
The failures look legit, I did not run it with -DallTests

spyrkob:
It just needs a change in CreateSnapshotTest it does not expect the new flag currently

prospero-cli/src/main/java/org/wildfly/prospero/cli/spi/CliProviderImpl.java
spyrkob: OK makes sense

spyrkob (approved):

PR #613, opened 2024-03-27
Hardcode channel properties

spyrkob:
Hardcode channel configuration in the feature pack


spyrkob:
Closing as this turns out to be not needed.

Issue #587, opened 2024-03-07
Allow execution of the jboss-eap-installation-manager.sh from inside of the `bin/` directory


TomasHofman:
```
$ ./jboss-eap-installation-manager.sh update list
ERROR: Path `/home/thofman/prospero/eap-8/bin/.` does not contain a server installation provisioned by the jboss-eap-installation-manager.
  Maybe you forgot to specify path to the installation (--dir)?
  Server installation needs to contain a `.galleon` folder and an `.installation/installer-channels.yaml` file.
```

It would be nice to be able to execute prospero directly from the `bin/` directory, because that's what I'm used to do :).

Maybe looking recursively into parent directories for the `.installation/` directory? Or just looking into the immediate parent dir if `.installation/` is not in the current dir?

spyrkob:
@TomasHofman I understand this applies when using prospero embedded inside Wildfly server right? If that's the case maybe we can change the script in the `wildfly-prospero-galleon-pack` itself to set the --dir to JBOSS_HOME?

TomasHofman:
@spyrkob I think that could work, if user is still able to override `--dir` by specifying it in his invocation...

spyrkob:
yeah that should be possible, if the argument is present the script wouldn't change it

spyrkob:
Looking at this again, the `prospero.sh|bat|ps1` script is shared between standalone and wildfly FPs, so I think it's better to make this change in `AbstractCommand#determineInstallationDirectory`.
I believe we can recursively check all parents until we find first folder with `.galleon`/.`.installation` subfolders, then validate it as currently.

parsharma:
I would like to work on this issue. Thanks

PR #637, opened 2024-04-16
587 Allow execution of the jboss-eap-installation-manager.sh from inside of the bin/ directory

parsharma:
Issue: https://github.com/wildfly-extras/prospero/issues/587

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java
@@ -68,13 +68,22 @@ public abstract class AbstractCommand implements Callable<Integer> {
     protected static Path determineInstallationDirectory(Optional<Path> directoryOption) throws ArgumentParsingException {
         Path installationDirectory = directoryOption.orElse(currentDir()).toAbsolutePath();
 
-        // Iterate through parent directories until a valid installation directory is found
-        while (!verifyDirectoryContainsInstallation(installationDirectory)) {
-            installationDirectory = installationDirectory.getParent();
-            if (installationDirectory == null) {
-                throw CliMessages.MESSAGES.invalidInstallationDirMaybeUseDirOption(installationDirectory);
+        // Check if the directory option was provided
+        if (directoryOption.isPresent()) {
+            // If --dir is present, verify the provided directory
+            if (!verifyDirectoryContainsInstallation(installationDirectory)) {
+                throw CliMessages.MESSAGES.invalidInstallationDir(installationDirectory);
+            }
+        } else {
+            // If --dir is not present, iterate through parent directories until a valid installation directory is found
+            while (!verifyDirectoryContainsInstallation(installationDirectory)) {
+                installationDirectory = installationDirectory.getParent();
+                if (installationDirectory == null) {
+                    throw CliMessages.MESSAGES.invalidInstallationDirMaybeUseDirOption(currentDir().toAbsolutePath());
+                }
             }
         }
+
         return installationDirectory;
     }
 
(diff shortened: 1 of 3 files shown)


parsharma:
@spyrkob why we are throwing two exception [here](https://github.com/wildfly-extras/prospero/blob/main/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java#L83-L85). It seems same to me. Could you please explain, may be I am missing something?

spyrkob:
The first case is meant to be triggered if the user didn't use `--dir` parameter, the second if they did use it. Admittedly  It should be refactored to be more explicit about it

TomasHofman:
@spyrkob should we test parent directories all the way up, as it is done currently, or should we only test single parent dir? (We would only expect the user to be in $JBOSS_HOME or $JBOSS_HOME/bin/ in the latter case.)

spyrkob:
I think all the way up. IMO it'd be a bit confusing why prospero can detect server when in JBOSS_HOME/modules but not in JBOSS_HOME/modules/base

parsharma:
Here I have updated the code but still one test case is failing. I am trying to figure out why it is failing.

spyrkob:
That actually brings a good point - I think this logic should only apply if user didn't use `--dir` parameter. If they used `--dir` or `--candidate-dir` Prospero should only check the folder that was selected

parsharma:
Sorry, I am not getting. Could you please elaborate?
Thanks

spyrkob:
Currently the user has two ways to execute prospero on a server installed in `${JBOSS_HOME}`: either navigate into `${JBOSS_HOME}` folder and execute prospero command e.g. `prospero update list`; or execute prospero command from any directory and point to the server folder using `--dir` parameter - e.g. `prospero update list --dir ${JBOSS_HOME}`.

The first case should work if the user navigates to any subfolder of the server - e.g. `${JBOSS_HOME}/bin`, `${JBOSS_HOME}/modules/base`, etc.

The later case should only work if the path specified by `--dir` points to the top-level folder of the server - i.e. `${JBOSS_HOME}`. Following would be therefore invalid: `prospero update list --dir ${JBOSS_HOME}/bin`

spyrkob:
@parsharma please add a unit test to verify the new logic

spyrkob (changes requested):
@parsharma please add a unit test to verify the new logic

TomasHofman:
@spyrkob when you get back this is ready for review.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java
```diff
@@ -67,24 +66,37 @@ public AbstractCommand(CliConsole console, ActionFactory actionFactory) {
...
-        File channelsFile = path.resolve(ProsperoMetadataUtils.METADATA_DIR)
-                .resolve(ProsperoMetadataUtils.INSTALLER_CHANNELS_FILE_NAME).toFile();
-        if (!dotGalleonDir.isDirectory() || !channelsFile.isFile()) {
-            if (currentDir().equals(path)) {
-                // if the path is the current directory, user may have forgotten to specify the --dir option
-                //   -> suggest to use --dir
-                throw CliMessages.MESSAGES.invalidInstallationDirMaybeUseDirOption(path);
-            } else {
-                throw CliMessages.MESSAGES.invalidInstallationDir(path);
+    static Path determineInstallationDirectory(Optional<Path> directoryOption, Path currentDir) throws ArgumentParsingException {
```
spyrkob, line 72: I think this can be private, testing can happen through the `determineInstallationDirectory` method or am I missing something?

spyrkob (changes requested):

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/AbstractCommand.java
TomasHofman: *This* is the `determineInstallationDirectory()` method. It now has two variants, one that only takes only `Optional<Path> directoryOption` parameter, and the second that takes `Optional<Path> directoryOption, Path currentDir` parameters.

The second parameter in the latter method is to make testing easier because otherwise it's not very well possible to set working directory inside a java process, that's why it's given as a method parameter.

So we need this latter method to be package private to make it visible for unit test.

spyrkob: Good point I was thinking about it backwards

PR #77, opened 2022-06-29
Validate at least single channel is provided, remove references to wildfly fpls

TomasHofman:
Cc @spyrkob

PR #668, opened 2024-05-14
Update the channel add command to accept channel as well as manifest

parsharma:
Fixes: https://github.com/wildfly-extras/prospero/issues/206

I've created a separate branch instead of using the old one. It would be helpful if you could review it @TomasHofman & @spyrkob. There are some remaining tasks, such as fixing test cases and completing the code. I want to ensure that I'm on the right track and not missing anything important.

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/ChannelUtils.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/ChannelUtils.java
@@ -1,0 +1,33 @@
+package org.wildfly.prospero.cli;
+
+import org.eclipse.aether.DefaultRepositorySystemSession;
+import org.eclipse.aether.RepositorySystem;
+import org.jboss.galleon.ProvisioningException;
+import org.wildfly.channel.Channel;
+import org.wildfly.channel.maven.VersionResolverFactory;
+import org.wildfly.prospero.api.MavenOptions;
+import org.wildfly.prospero.api.ProvisioningDefinition;
+import org.wildfly.prospero.api.exceptions.ChannelDefinitionException;
+import org.wildfly.prospero.api.exceptions.NoChannelException;
+import org.wildfly.prospero.wfchannel.MavenSessionManager;
+
+import java.util.List;
+
+public class ChannelUtils {
+
+    public static VersionResolverFactory createVersionResolverFactory(MavenSessionManager mavenSessionManager) {
+        final RepositorySystem repositorySystem = mavenSessionManager.newRepositorySystem();
+        final DefaultRepositorySystemSession repositorySystemSession = mavenSessionManager.newRepositorySystemSession(
+                repositorySystem);
+        return new VersionResolverFactory(repositorySystem, repositorySystemSession);
+    }
+
+    public static List<Channel> resolveChannels(ProvisioningDefinition provisioningDefinition, MavenOptions mavenOptions)
+            throws ArgumentParsingException, ProvisioningException, NoChannelException, ChannelDefinitionException {
+        final MavenSessionManager mavenSessionManager = new MavenSessionManager(mavenOptions);
+        final VersionResolverFactory versionResolverFactory = createVersionResolverFactory(mavenSessionManager);
+        final List<Channel> channels = provisioningDefinition.resolveChannels(versionResolverFactory);
+        return channels;
+    }
+
+}
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
@@ -19,20 +19,24 @@ package org.wildfly.prospero.cli.commands.channel;
 
 import org.wildfly.channel.Channel;
 import org.wildfly.channel.ChannelManifestCoordinate;
+import org.wildfly.channel.ChannelMapper;
 import org.wildfly.channel.Repository;
+import org.wildfly.channel.InvalidChannelMetadataException;
 import org.wildfly.prospero.actions.MetadataAction;
 import org.wildfly.prospero.api.ArtifactUtils;
 import org.wildfly.prospero.api.RepositoryUtils;
-import org.wildfly.prospero.cli.ActionFactory;
+import org.wildfly.prospero.api.TemporaryFilesManager;
 import org.wildfly.prospero.cli.CliConsole;
+import org.wildfly.prospero.cli.ActionFactory;
 import org.wildfly.prospero.cli.CliMessages;
 import org.wildfly.prospero.cli.RepositoryDefinition;
 import org.wildfly.prospero.cli.ReturnCodes;
-import org.wildfly.prospero.api.TemporaryFilesManager;
 import org.wildfly.prospero.cli.commands.AbstractCommand;
 import org.wildfly.prospero.cli.commands.CliConstants;
 import picocli.CommandLine;
 
+import java.nio.file.Files;
+import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Optional;
@@ -43,14 +47,27 @@ public class ChannelAddCommand extends AbstractCommand {
     @CommandLine.Option(names = CliConstants.CHANNEL_NAME, required = true)
     private String channelName;
 
-    @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST, required = true)
-    private String gavUrlOrPath;
+    @CommandLine.ArgGroup(exclusive = true, multiplicity = "1")
+    ChannelParamsGroup channelOptions;
 
     @CommandLine.Option(names = CliConstants.DIR)
     private Optional<Path> directory;
 
-    @CommandLine.Option(names = CliConstants.REPOSITORIES, split = ",", paramLabel = CliConstants.REPO_URL, required = true)
-    private List<String> repositoryDefs;
+    static class ChannelParamsGroup {
+        @CommandLine.ArgGroup(exclusive = false, multiplicity = "1")
+        private ChannelGroup channelGroup;
+
+        @CommandLine.Option(names = CliConstants.CHANNEL)
+        private Path channelLocation;
+    }
+
+    static class ChannelGroup {
+        @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST, required = true)
+        private String manifestLocation;
+
+        @CommandLine.Option(names = CliConstants.REPOSITORIES, split = ",", paramLabel = CliConstants.REPO_URL, required = true)
+        private List<String> repositoryDefs;
+    }
 
     public ChannelAddCommand(CliConsole console, ActionFactory actionFactory) {
         super(console, actionFactory);
@@ -59,20 +76,38 @@ public class ChannelAddCommand extends AbstractCommand {
     @Override
     public Integer call() throws Exception {
         Path installationDirectory = determineInstallationDirectory(directory);
-
-        console.println(CliMessages.MESSAGES.subscribeChannel(installationDirectory, channelName));
-
-        ChannelManifestCoordinate manifest = ArtifactUtils.manifestCoordFromString(gavUrlOrPath);
-
-        try (TemporaryFilesManager temporaryFiles = TemporaryFilesManager.getInstance()) {
-            final List<Repository> repositories = RepositoryUtils.unzipArchives(RepositoryDefinition.from(repositoryDefs), temporaryFiles);
-
-            Channel channel = new Channel(channelName, null, null, repositories, manifest, null, null);
-            try (MetadataAction metadataAction = actionFactory.metadataActions(installationDirectory)) {
-                metadataAction.addChannel(channel);
+        List<Channel> channels = null;
+        if (channelOptions.channelLocation != null) {
+            try{
+                channels = ChannelMapper.fromString(Files.readString(Path.of(channelOptions.channelLocation.toString())));
+            }catch (InvalidChannelMetadataException e){
+                    throw CliMessages.MESSAGES.invalidManifest(e.getMessage());
+            }catch (NoSuchFileException e){
+                throw CliMessages.MESSAGES.fileNotExists(e.getMessage());
             }
-        }
 
+            Channel channel = channels.get(0);
+            channels.set(0, new Channel(channelName, channel.getDescription(), channel.getVendor(), channel.getRepositories(), channel.getManifestCoordinate(), channel.getBlocklistCoordinate(), channel.getNoStreamStrategy()));
+            if (channels.size() > 1) {
+                throw new RuntimeException(CliMessages.MESSAGES.sizeOfChannel());
+            }
+        } else if (!(channelOptions.channelGroup.manifestLocation.isEmpty() || channelOptions.channelGroup.repositoryDefs.isEmpty())) {
+            console.println(CliMessages.MESSAGES.subscribeChannel(installationDirectory, channelName));
+
+            ChannelManifestCoordinate manifest = ArtifactUtils.manifestCoordFromString(channelOptions.channelGroup.manifestLocation);
+            try (TemporaryFilesManager temporaryFiles = TemporaryFilesManager.getInstance()) {
+                final List<Repository> repositories = RepositoryUtils.unzipArchives(RepositoryDefinition.from(channelOptions.channelGroup.repositoryDefs), temporaryFiles);
+                channels = List.of(new Channel(channelName, null, null, repositories, manifest, null, null));
+
+            }
+        } else {
+            throw new RuntimeException(CliMessages.MESSAGES.invalidArgument());
+        }
+        try (MetadataAction metadataAction = actionFactory.metadataActions(installationDirectory)) {
+            Channel channel = channels.get(0);
+                metadataAction.addChannel(channel);
+
+        }
         console.println(CliMessages.MESSAGES.channelAdded(channelName));
         return ReturnCodes.SUCCESS;
 
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
@@ -22,10 +22,13 @@ import org.wildfly.channel.ChannelManifestCoordinate;
 import org.wildfly.channel.ChannelMapper;
 import org.wildfly.channel.Repository;
 import org.wildfly.channel.InvalidChannelMetadataException;
+import org.wildfly.prospero.ProsperoLogger;
 import org.wildfly.prospero.actions.MetadataAction;
 import org.wildfly.prospero.api.ArtifactUtils;
 import org.wildfly.prospero.api.RepositoryUtils;
 import org.wildfly.prospero.api.TemporaryFilesManager;
+import org.wildfly.prospero.api.exceptions.OperationException;
+import org.wildfly.prospero.cli.ArgumentParsingException;
 import org.wildfly.prospero.cli.CliConsole;
 import org.wildfly.prospero.cli.ActionFactory;
 import org.wildfly.prospero.cli.CliMessages;
@@ -35,8 +38,8 @@ import org.wildfly.prospero.cli.commands.AbstractCommand;
 import org.wildfly.prospero.cli.commands.CliConstants;
 import picocli.CommandLine;
 
+import java.io.IOException;
 import java.nio.file.Files;
-import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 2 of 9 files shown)


prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -20,61 +20,95 @@
...
 public class ChannelAddCommand extends AbstractCommand {
 
-    @CommandLine.Option(names = CliConstants.CHANNEL_NAME, required = true)
+    @CommandLine.Option(names = CliConstants.CHANNEL_NAME)
     private String channelName;
 
-    @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST, required = true)
-    private String gavUrlOrPath;
+    @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST)
+    private String manifestLocation;
```
spyrkob, line 56: `channelName` should always be requires, even if the channelFile is provided. 

Also the command syntax should enforce that either `channelName`, `repositories` and `manifestLocation` is present or `channelLocation` is present. I thing grouping the arguments using `@CommandLine.ArgGroup` should be able to achieve that.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -20,61 +20,95 @@
...
     public ChannelAddCommand(CliConsole console, ActionFactory actionFactory) {
         super(console, actionFactory);
     }
 
     @Override
     public Integer call() throws Exception {
         Path installationDirectory = determineInstallationDirectory(directory);
+        List<Channel> channels = null;
+        if (!channelLocation.isEmpty()) {
+            final ChannelCoordinate coord = ArtifactUtils.channelCoordFromString(channelLocation);
```
spyrkob, line 77: I'm not sure if this is needed. If the location is simple URL or a path to the channel file, we don't need the whole ArtifactUtils.channelCoordFromString/ChannelUtils.resolveChannels thing, it should be enough to read the file using ChannelMapper.

Handling a case where the location is a GA[V] is more complex, but I'm not sure if that's something we want to be doing.

prospero-cli/test-path/org/test/test/resolver-status.properties
spyrkob, line 1: Is that included by accident?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -20,61 +20,95 @@
...
+            checkFileExists(coord.getUrl(), channelLocation);
+            final ProvisioningDefinition provisioningDefinition = ProvisioningDefinition.builder()
+                    .setChannelCoordinates(channelLocation)
+                    .build();
+            channels = ChannelUtils.resolveChannels(provisioningDefinition, MavenOptions.DEFAULT_OPTIONS);
```
spyrkob, line 82: The action has to ensure the all the channels define non-empty name attributes

spyrkob (changes requested):
looks good, a few comments in the code

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
parsharma: Do you mean all three channelName, repo and manifestLocation should be present? If channelLocation is not present

prospero-cli/test-path/org/test/test/resolver-status.properties
parsharma: Yes, I'll remove it.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
spyrkob: yes, exactly

prospero-cli/src/main/java/org/wildfly/prospero/cli/ChannelUtils.java
```diff
@@ -0,0 +1,33 @@
...
+public class ChannelUtils {
+
+    public static VersionResolverFactory createVersionResolverFactory(MavenSessionManager mavenSessionManager) {
+        final RepositorySystem repositorySystem = mavenSessionManager.newRepositorySystem();
+        final DefaultRepositorySystemSession repositorySystemSession = mavenSessionManager.newRepositorySystemSession(
+                repositorySystem);
+        return new VersionResolverFactory(repositorySystem, repositorySystemSession);
+    }
+
+    public static List<Channel> resolveChannels(ProvisioningDefinition provisioningDefinition, MavenOptions mavenOptions)
```
parsharma, line 25: @spyrkob @TomasHofman Do you think we'll need the ChannelUtils class in the future?

TomasHofman: I think we could even keep this, but whatever @spyrkob likes is fine.

spyrkob: Yeah keeping it is fine, the logic doesn't really belong in in AbstractInstallCommand IMO

parsharma:
One build ubuntu-latest-pmd is failing but it is not failing in local.  Can we rerun this build if possible?

TomasHofman:
Did you try locally the same command the job is running?
```
mvn pmd:aggregate-pmd-check -DskipTests
```

TomasHofman:
pmd is sort of like checkstyle, so it looks like legitimate failure...

parsharma:
@TomasHofman yes now it is failing after running PMD command. Error is not clear. Where should I check?

prospero-cli/src/main/java/org/wildfly/prospero/cli/ChannelUtils.java
```diff
@@ -0,0 +1,33 @@
...
+            throws ArgumentParsingException, ProvisioningException, NoChannelException, ChannelDefinitionException {
+        final MavenSessionManager mavenSessionManager = new MavenSessionManager(mavenOptions);
+        final VersionResolverFactory versionResolverFactory = ChannelUtils.createVersionResolverFactory(mavenSessionManager);
```
parsharma, line 28: I am calling ChannelUtils.createVersion.... here PMD check is failing.

spyrkob:
just FYI the gitlab action prints the summary of PMD check in a "Display report" step after running the tests, you can see the report there

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -43,36 +45,55 @@ public class ChannelAddCommand extends AbstractCommand {
     @CommandLine.Option(names = CliConstants.CHANNEL_NAME, required = true)
     private String channelName;
 
-    @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST, required = true)
-    private String gavUrlOrPath;
+    @CommandLine.ArgGroup(exclusive = false)
+    ChannelGroup channelGroup;
+
+    static class ChannelGroup {
```
spyrkob, line 51: Can you try to make them exclusive with channelLocation? I think it should be possible using a nested ArgGroup.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -43,36 +45,55 @@ public class ChannelAddCommand extends AbstractCommand {
...
+            ChannelManifestCoordinate manifest = ArtifactUtils.manifestCoordFromString(channelGroup.manifestLocation);
+            try (TemporaryFilesManager temporaryFiles = TemporaryFilesManager.getInstance()) {
+                final List<Repository> repositories = RepositoryUtils.unzipArchives(RepositoryDefinition.from(channelGroup.repositoryDefs), temporaryFiles);
+                channels = List.of(new Channel(channelName, null, null, repositories, manifest, null, null));
 
-            Channel channel = new Channel(channelName, null, null, repositories, manifest, null, null);
-            try (MetadataAction metadataAction = actionFactory.metadataActions(installationDirectory)) {
+            }
+        } else {
+            return ReturnCodes.INVALID_ARGUMENTS;
```
spyrkob, line 90: Please add error message or even better throw an exception with the message.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -43,36 +45,55 @@ public class ChannelAddCommand extends AbstractCommand {
...
+        }
+        try (MetadataAction metadataAction = actionFactory.metadataActions(installationDirectory)) {
+            for (Channel channel : channels) {
```
spyrkob, line 93: Since it's only possible to add one channel from file ( line #77), you don't need the loop here. There will always be just a single channel

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -43,36 +45,55 @@ public class ChannelAddCommand extends AbstractCommand {
...
     public Integer call() throws Exception {
         Path installationDirectory = determineInstallationDirectory(directory);
+        List<Channel> channels = null;
+        if (channelLocation != null && !channelLocation.isEmpty()) {
+            String yaml = Files.readString(Path.of(channelLocation));
+            channels = ChannelMapper.fromString(yaml);
+            Channel channel = channels.get(0);
+            channels.set(0, new Channel(channelName, channel.getDescription(), channel.getVendor(), channel.getRepositories(), channel.getManifestCoordinate(), channel.getBlocklistCoordinate(), channel.getNoStreamStrategy()));
+            if (channels.size() > 1) {
+                throw new RuntimeException("More than one channel found in the list.");
```
spyrkob, line 78: This needs to a translated text - it should be added to CliMessages

spyrkob (changes requested):
Please add unit tests to `ChannelCommandTest`

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -43,36 +45,55 @@ public class ChannelAddCommand extends AbstractCommand {
...
+    ChannelGroup channelGroup;
+
+    static class ChannelGroup {
+        @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST)
+        private String manifestLocation;
+
+        @CommandLine.Option(names = CliConstants.REPOSITORIES, split = ",", paramLabel = CliConstants.REPO_URL)
+        private List<String> repositoryDefs;
+    }
+    @CommandLine.Option(names = CliConstants.CHANNEL)
```
spyrkob, line 58: The default `--channel` parameter description (prospero channel add --help) doesn't work for this, we need to change it. See UsageMessages.properties, you'll need to add keys like `${prospero.dist.name}.channel.add.channel.0`, `${prospero.dist.name}.channel.add.channel.1`, `${prospero.dist.name}.channel.add.channel.2`

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
```diff
@@ -43,36 +45,55 @@ public class ChannelAddCommand extends AbstractCommand {
...
     private String channelName;
 
-    @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST, required = true)
-    private String gavUrlOrPath;
+    @CommandLine.ArgGroup(exclusive = false)
+    ChannelGroup channelGroup;
+
+    static class ChannelGroup {
+        @CommandLine.Option(names = CliConstants.CHANNEL_MANIFEST)
+        private String manifestLocation;
```
spyrkob, line 53: This can be a Path rather then a String

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/ChannelCommandTest.java
```diff
@@ -49,6 +51,7 @@ public class ChannelCommandTest extends AbstractConsoleTest {
     private static final MavenCoordinate GA = new MavenCoordinate("g", "a", null);
     private static final MavenCoordinate GAV = new MavenCoordinate("g", "a", "v");
     private static final String URL = "file:/a:b";
+    private final URL testChannelURL = getClass().getClassLoader().getResource("test-channel.yaml");
```
parsharma, line 54: @spyrkob I am using static yaml file as it is reusable and seems easier. WDT?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
TomasHofman: I checking the windows failures and have a suspicion that this change to Path causes problems.

```
Invalid value for option '--manifest': cannot convert 'org.test:test' to interface java.nio.file.Path (java.nio.file.InvalidPathException: Illegal char <:> at index 8: org.test:test)
```
https://github.com/wildfly-extras/prospero/actions/runs/9114495539/job/25058577001?pr=668

I think the core issue is that manifestLocation should remain a String, because it may not be a path, it can still be a maven GAV..?

spyrkob: ugh sorry, my bad that was meant to be a comment about `channelLocation` not `manifestLocation`

TomasHofman: And if that is true, and --manifest can be a GAV, then the same rules should probably apply to --channel (alias channelLocation).

Currently I can see channelLocation being still a String, but later in the code being treated as path:

```
ChannelAddCommand
...
public Integer call() throws Exception {
        Path installationDirectory = determineInstallationDirectory(directory);
        List<Channel> channels = null;
        if (channelOptions.channelLocation != null && !channelOptions.channelLocation.isEmpty()) {
            String yaml = Files.readString(Path.of(channelOptions.channelLocation));
            channels = ChannelMapper.fromString(yaml);
            Channel channel = channels.get(0);
            channels.set(0, new Channel(channelName, channel.getDescription(), channel.getVendor(), channel.getRepositories(), channel.getManifestCoordinate(), channel.getBlocklistCoordinate(), channel.getNoStreamStrategy()));
            if (channels.size() > 1) {
                throw new RuntimeException(CliMessages.MESSAGES.sizeOfChannel());
            }
        } else if ...
```

and this is I think caused by us giving wrong instructions to @parsharma. She originally used the resolver to obtain the channel instance, which IMO was correct, if channelLocation can be a GAV. Then we told her to use ChannelMapper which only works for local paths.

WDYT @spyrkob ?

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/ChannelCommandTest.java
spyrkob: You can simply use `final Path path = MetadataTestUtils.prepareChannel("manifest.yaml");` instead. I'd rather avoid adding more channel definitions that need to be maintained manually.

Also, as far as I can tell, this field is not used at all.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/channel/ChannelAddCommand.java
spyrkob: Yes, for simplicity I'd rather go with local channels only - at least initially. When resolving channels from GAV we'd need to add options like offline mode and repositories (which gets even more confusing with existing repositories parameter).

At very least I think this should be two separate parameters - maybe something like `--channel-file` and `--channel-gav` - it's much easier to explain it in documentation this way (and less buggy to handle in code).

TomasHofman: So for @parsharma , this means:

1. change manifestLocation back to String :),
2. change channelLocation to Path, and leave the rest as is.

TomasHofman: Considerations out-of-scope for this PR:

I agree the repositories parameter is bit overloaded. This may be similar even in the install command however. One thing is which repository the channel / manifest should be looked up, another where the channel artifacts should be resolved from.

The split to --channel-file, --channel-gav can make sense, but if it's ever done if should probably be done everywhere. Which means big change for users. Having just single command take different kind of parameters from all the other commands is weird.

Maybe refactoring the resolution logic for the current universal gav/url/path parameters, so they are easier to reuse.

prospero-cli/src/main/resources/UsageMessages.properties
```diff
@@ -318,6 +321,8 @@ prospero.channels.added=Channel '%s' added.
 prospero.channels.remove.header=Unsubscribing %s from channel %s%n
 prospero.channels.removed=Channel '%s' removed.
 prospero.channels.error.notfound=The requested channel doesn't exist
+prospero.channels.error.morethanonechannel.found=More than one channel found
+prospero.channels.error.arguments.incorrect=Passing invalid arguments
```
spyrkob, line 325: @parsharma the error message should let the user know what is the problem. In this case the problem is the command requires a channel definition (either via `--channel` or other parameters).

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/ChannelCommandTest.java
```diff
@@ -277,6 +280,15 @@ public void testRemoveEmptyChannel() {
                 CliConstants.DIR, dir.toString());
         Assert.assertEquals(ReturnCodes.INVALID_ARGUMENTS, exitCode);
     }
+    @Test
```
spyrkob, line 283: Other tests:
 * error when `--channel` is used together with `--manifest` or `--repositories`
 * error when `--channel` is used with non-existing file
 * error when `--channel` is used with invalid channel file
 * verify that the recorded channel has the same content (minus name) as the one passed in

spyrkob (changes requested):

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/ChannelCommandTest.java
```diff
@@ -277,6 +283,59 @@ public void testRemoveEmptyChannel() {
...
+                "    repositories:",
+                "      - id: test-repo",
+                "        url: http://invalid-url.org/repo"
+        );
+        Files.write(InvalidChannelFile.toPath(), yamlContent);
+        exitCode = commandLine.execute(CliConstants.Commands.CHANNEL, CliConstants.Commands.ADD,
+                CliConstants.DIR, dir.toString(),
+                CliConstants.CHANNEL_NAME, "channel-2",
+                CliConstants.CHANNEL, InvalidChannelFile.getAbsolutePath());
+        Assert.assertEquals(ReturnCodes.PROCESSING_ERROR, exitCode);
```
parsharma, line 337: @spyrkob I am not sure why exit code is coming 2 here. Could you please take a look?
Thanks

spyrkob: @parsharma do you mean why it's returning PROCESSING_ERROR? Either code is fine in this case. Generally INVALID_ARGUMENT is related to syntax -e.g. missing argument, etc while PROCESSING_ERROR refers to issue executing command.

However when running this example in command line I get:
```
org.wildfly.channel.InvalidChannelMetadataException: Invalid Manifest
caused by:
 The manifest does not specify a schemaVersion.

	at org.wildfly.channel.ChannelMapper.getSchema(ChannelMapper.java:80)
	at org.wildfly.channel.ChannelMapper.validateString(ChannelMapper.java:146)
	at org.wildfly.channel.ChannelMapper.fromString(ChannelMapper.java:125)
	at org.wildfly.prospero.cli.commands.channel.ChannelAddCommand.call(ChannelAddCommand.java:80)
	at org.wildfly.prospero.cli.commands.channel.ChannelAddCommand.call(ChannelAddCommand.java:42)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2045)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
	at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2465)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2457)
	at picocli.CommandLine$RunLast.handle(CommandLine.java:2419)
	at picocli.CommandLine$AbstractParseResultHandler.execute(CommandLine.java:2277)
	at picocli.CommandLine$RunLast.execute(CommandLine.java:2421)
	at picocli.CommandLine.execute(CommandLine.java:2174)
	at org.wildfly.prospero.cli.CliMain.execute(CliMain.java:132)
	at org.wildfly.prospero.cli.CliMain.main(CliMain.java:60)
```

The program shouldn't show exception and stack trace to the user, rather a message describing the problem. `InstallCommand` handles similar case for `--channel` argument

parsharma:
@spyrkob PR is ready for review.

PR #679, opened 2024-05-22
Avoid closing Provisioning object if it was not used

spyrkob:
Running `prospero install --profile wildfly` results in logging of an error "Error releasing classloader Releasing usage of core 6.0.1.Final although no usage".

The server is generated correctly, the error is related to Galleon 6.0.0 upgrade.


yersan (approved):

yersan:
This fixed the errors I was getting provisioning a server

PR #689, opened 2024-06-05
[JBEAP-27255] Return the status of java execution in bat script

spyrkob:
Upstream: #688
Issue: https://issues.redhat.com/browse/JBEAP-27255


spyrkob:
superceeded by #707

PR #701, opened 2024-06-24
Handle installation directory being a symbolic link

spyrkob:
When specified Wildfly installation directory (JBOSS_HOME) is a symbolic link, applying update fails with "ERROR: java.nio.file.FileAlreadyExistsException: <specified-path-of-JBOSS_HOME-directory>".

Actually, it fails only when applying updates to files directly under the link (e.g. copying jboss-modules.jar), so other files under modules directory are appearantly updated succeessfully before the ERROR message.


iweiss:
LGTM

PR #702, opened 2024-06-25
[JBEAP-27356] Handle installation directory being a symbolic link

spyrkob:
Upstream: #701
Issue: https://issues.redhat.com/browse/JBEAP-27356

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/actions/InstallFolderUtils.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/actions/InstallFolderUtils.java
@@ -19,6 +19,7 @@ package org.wildfly.prospero.actions;
 
 import org.wildfly.prospero.ProsperoLogger;
 
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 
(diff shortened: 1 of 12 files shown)


prospero-common/src/main/java/org/wildfly/prospero/actions/InstallFolderUtils.java
```diff
@@ -40,6 +41,25 @@ static void verifyIsEmptyDir(Path directory) {
...
+
+        while (path != null && !(Files.exists(path) && Files.isSymbolicLink(path))) {
+            path = path.getParent();
+        }
+
+        if (path != null) {
+            final Path relativized = path.relativize(symlink);
+            try {
+                return path.toRealPath().resolve(relativized);
+            } catch (IOException e) {
```
TomasHofman, line 55: Not sure, but should there be another check if the path is really a symlink, before doing the `path.relativize()` and the rest?

It seems like we call this method always to sanitize the input.

T double check - IIUC the issue only is if the last segment of the path is a symlink. A symlink in the middle of the path does not cause an issue..?

prospero-common/src/main/java/org/wildfly/prospero/actions/InstallFolderUtils.java
```diff
@@ -40,6 +41,25 @@ static void verifyIsEmptyDir(Path directory) {
         }
     }
 
+    static Path toRealPath(Path symlink) {
+        Path path = symlink;
+
+        while (path != null && !(Files.exists(path) && Files.isSymbolicLink(path))) {
+            path = path.getParent();
+        }
```
TomasHofman, line 49: To me it looks bit more complex than needed, though I just may be missing something. What is the reason for this cycle?

I'm trying to reproduce the issue, following fails with the FileAlreadyExistsException:

```
Path symbolicLink = Files.createSymbolicLink(Path.of("/tmp/symlink"), Path.of("/tmp"));
Files.createDirectories(Path.of("/tmp/symlink"));
```

But following is OK:

```
Path symbolicLink = Files.createSymbolicLink(Path.of("/tmp/symlink"), Path.of("/tmp"));
Files.createDirectories(Path.of("/tmp/symlink/a")); // deeper paths are OK too
```

So to me it looks like we only need to check the complete path as given.

prospero-common/src/main/java/org/wildfly/prospero/actions/InstallFolderUtils.java
TomasHofman: You can disregard this comment :).

prospero-common/src/main/java/org/wildfly/prospero/actions/InstallFolderUtils.java
TomasHofman: It does seem to work though. I think perhaps it could be simplified but it is fine as it is IMO. :+1:

TomasHofman: BTW I got the cycle meaning so no need to elaborate :).

What I meant by simplifying is doing just:

```
if (Files.isSymbolicLink(path)) {
    return path.toRealPath();
}
return path;
```

but that's just for consideration...

TomasHofman (approved):

prospero-common/src/main/java/org/wildfly/prospero/actions/InstallFolderUtils.java
spyrkob: Thanks for review @TomasHofman.

Yes the issue is only reproducible if the top level dir is a symlink, so it would be enough to use the simplified code. 

In this case I wanted to make sure the Path passed further on is consistent regardless if user passed the symlink or a subdirectory of symlink. In my mind it makes it easier to reason about further execution instead of keeping the two possibilities in mind.

If you think it's better to use the simpler approach, I don't mind changing it. Either way I should add this explanation to code :)

TomasHofman: I think either way is fine, it just took me a bit to decipher that sanitizing method.

PR #705, opened 2024-06-25
Add a build-image command to build a container image from a WildFly installation

jmesnil:
Add JIB as a dependency to build the image (docker or podman are not required).

* --name -> required parameter to specify the name of the container image
* --dir -> optional parameter to specify the installation directory. if absent, search in the parents directory for an installation
* --runtime-version -> optional parameter to specify the tag of the WildFly runtime image (default is latest)
* --user / --password -> optional credentials to push the image to a container registry

### DIFF
--- a/prospero-cli/pom.xml
+++ b/prospero-cli/pom.xml
@@ -13,6 +13,12 @@
     <packaging>jar</packaging>
 
     <dependencies>
+
+        <dependency>
+            <groupId>com.google.cloud.tools</groupId>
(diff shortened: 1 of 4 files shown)


spyrkob:
Nice :) 

Do you see more cloud/image related tasks added to prospero? If so it might make sense to create a sub-command to organise them.

jmesnil:
I did not thoroughly thought about it but you're making a good point `image build` instead of `build-image` would be more consistent with other commands/subcommands

PR #708, opened 2024-06-28
Rollback changes in case of unsucessful apply stage

spyrkob:
Before applying changes to a server, backup the state of the server in a `<SERVER_HOME>.update.old` folder. The backup is done using hardlinks if the file system allows it to save disk space. 

If the update fails due to IOException, we try and restore the server state. If the update is interrupted, either by the user or a program crash, the original state will be available for the user to perform manual restore.

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyStageBackup.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyStageBackup.java
@@ -7,7 +7,6 @@ import org.wildfly.prospero.ProsperoLogger;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
-import java.io.UnsupportedEncodingException;
 import java.nio.file.FileVisitResult;
 import java.nio.file.Files;
 import java.nio.file.Path;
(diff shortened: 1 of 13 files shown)


spyrkob:
@TomasHofman if you have a moment could you review this PR please?

TomasHofman:
Sure, will check.

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyStageBackup.java
```diff
@@ -0,0 +1,216 @@
...
+                        // when using hardlinks, we need to remove the file and copy the new one in it's place otherwise both would be changed.
+                        // the git folder is manipulated by jgit and we can't control how it operates on the files
+                        // therefore we need to copy the files upfront rather than hardlinking them
+                        Files.copy(file, backupRoot.resolve(relative));
+                    } else {
+                        // we try to use hardlinks instead of copy to save disk space
+                        // fallback on copy if Filesystem doesn't support hardlinks
+                        try {
+                            Files.createLink(backupRoot.resolve(relative), file);
+                        } catch (UnsupportedEncodingException e) {
```
TomasHofman, line 98: Is that the exception that's gonna be thrown when unable to create a hardlink? I would expect UnsupportedOperationException.

TomasHofman: We may try to check explicitly if the paths are on the same filesystem and revert to copy if they are not, to avoid the exception. It would not simplify the code, just avoided the exception in one predictable case. No big deal though.

spyrkob: good catch, thanks. That should be `UnsupportedOperationException`. 

> We may try to check explicitly if the paths are on the same filesystem.

That would assume the installation of WildFly itself is split between different filesystems, Is that something we support/anticipate?

TomasHofman: Hmm, probably not :+1:

TomasHofman (approved):
Looks OK to me.

Issue #704, opened 2024-06-25
Use jackson-databind library in WildFly modules in the WildFly feature pack


spyrkob:
Instead of adding the jackson-databind to prospero-dep module, use `com.fasterxml.jackson.dataformat.jackson-dataformat-yaml` wildfly-ee module.

parsharma:
@spyrkob can I take it?

spyrkob:
@parsharma  sure

TomasHofman:
@spyrkob , how would standalone prospero installation inherit the the module from wildfly-ee?

TomasHofman:
Aha - there are two prospero-dep modules, so this would only be done in the wildfly-galleon-pack module?

TomasHofman:
The com.fasterxml.jackson.dataformat.jackson-dataformat-yaml is only available in XP5, isn't that a problem?

PR #710, opened 2024-07-01
Use jackson-databind library in WildFly modules in the WildFly feature pack #704

parsharma:
Issue: https://github.com/wildfly-extras/prospero/issues/704


parsharma:
We need to update the dependency to WildFly 33 because our CI will not pass without the latest dependency.
<version.org.wildfly.galleon-pack>32.0.1.Final</version.org.wildfly.galleon-pack>
<version.org.wildfly.galleon-pack>33.0.0.Beta1-SNAPSHOT</version.org.wildfly.galleon-pack>

spyrkob:
@parsharma WildFly 33.0.0.Final ha been released and is upgraded in prospero main, please rebase your PR.

Note apart from adding the dependency, you need to move the `jackson-dataformat-yaml` artifact from the common `org.jboss.prospero` module to the standalone's `org.jboss.prospero-dep` module

spyrkob:
Ah right, the WildflyFpTest uses a hardcoded wildfly-28 manifest because back then WildFly didn't publish manifests... it will have to be updated to use the latest published WildFly manifest

PR #720, opened 2024-07-11
[JBEAP-27509] extending history information

spyrkob:
Add current manifest versions to the git commit so that they can be displayed in history operations

== Overview
The channel manifest 1.1.0 adds a notion of `logicalVersion` to separate the version of manifest from the Maven coordinate. This change will consume the new field falling back to the Maven coordinate (physical version) if the logicalVersion is not present.
Additionally, the PR changes how the manifest version information is stored. Previously the version string visible in history information was constructed during update and stored. This make it hard for API clients that would like to display the versions of individual manifests.

== Expected results
1.  History operation on servers installed with old versions of Prospero works as previously
2. A newly provisioned server's history contains individual manifest's version information rather than the constructed version message
3. Installation-manger-api returns manifest version information for each history state including logical and physical version. If the commit was made before this changes, API should return commit information instead.

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/api/SavedState.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/api/SavedState.java
@@ -18,10 +18,18 @@
 package org.wildfly.prospero.api;
 
 import java.time.Instant;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 import java.util.Locale;
 import java.util.Objects;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 public class SavedState {
+
+    private Set<Version> manifestVersions = Collections.emptySet();
 
     public enum Type { UPDATE, INSTALL, ROLLBACK, CONFIG_CHANGE,
         /**
@@ -50,11 +58,21 @@ public class SavedState {
     private Type type;
     private String msg = "";
 
+    @Deprecated
     public SavedState(String hash, Instant timestamp, Type type, String msg) {
         this.hash = hash;
         this.timestamp = timestamp;
         this.type = type;
         this.msg = msg;
+    }
+
+    public SavedState(String hash, Instant timestamp, Type type, String msg, List<Version> manifestVersions) {
+        this.hash = hash;
+        this.timestamp = timestamp;
+        this.type = type;
+        this.msg = msg;
+        // sort the manifests alphabetically so that the order of manifests in display will be consistent
+        this.manifestVersions = new TreeSet<>(manifestVersions);
     }
 
     public SavedState(String hash) {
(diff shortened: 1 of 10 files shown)


spyrkob:
@TomasHofman could you review this when you have a moment?

prospero-common/src/main/java/org/wildfly/prospero/api/SavedState.java
```diff
@@ -79,20 +93,96 @@ public String getMsg() {
...
+        return String.format("[%s] %s - %s %s", hash, timestamp.toString(), type.toString().toLowerCase(Locale.ROOT), msg);
     }
 
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         SavedState that = (SavedState) o;
-        return Objects.equals(hash, that.hash) && Objects.equals(timestamp, that.timestamp) && type == that.type;
+        return Objects.equals(manifestVersions, that.manifestVersions) && Objects.equals(hash, that.hash) && Objects.equals(timestamp, that.timestamp) && type == that.type && Objects.equals(msg, that.msg);
```
TomasHofman, line 116: It's probably not very important, but since the manifest versions are involved in the equals method, should the collection type for versions be Set, rather than List?

(Thinking purely about ordering, which I assume is generally undefined - version collections should be equal if they contain the same elements, irrespective of order. I think Set would facilitate that.)

TomasHofman:
To double check if this was intended, when I run the new history command against an old installation, I think it delegates to the old impl and shows something like this:

```
[thofman@rama prospero]$ ./prospero history --dir wf
[d0efe823] 2024-07-12T14:37:36Z - update [file:/home/thofman/Projects/prospero/manifest.yaml::a4ff586c9fce21976f8596c53d6247e200d22cc6]
[71420b8d] 2024-07-12T14:24:54Z - install [org.wildfly.channels:wildfly::33.0.0.Beta1]
```

It shows "install [G:A::version]" at the first record.

Running the command against a newly created installation, it doesn't show the GA anymore, just the version:

```
[thofman@rama prospero]$ ./prospero history --dir wf3
[7822a809] 2024-10-16T11:27:47Z - install [34.0.0.Beta1]
```

TomasHofman:
High level question: since the original impl of the history command also shows versions of manifests, what is the intended improvement here?

spyrkob:
Thanks @TomasHofman I updated the description to explain the changes. Let me know if that clarifies it.

> Running the command against a newly created installation, it doesn't show the GA anymore, just the version:
That's my mistake, I'll fix that thanks

> To double check if this was intended, when I run the new history command against an old installation, I think it delegates to the old impl and shows something like this
Correct, the old history states (made before this change) should be displayed as before

TomasHofman:
OK, thanks for the explanation, I noticed the new schema version by accident but didn't know the details. Now it makes sense to me :+1: .

TomasHofman (approved):

spyrkob:
Thanks @TomasHofman 👍

PR #735, opened 2024-07-22
[#726] Allow multiple manifests in the install command

### LINKED ISSUE #726: The `--manifest` parameter doesn't accept list of coordinates
The `--channel` parameter accepts list of coordinates, which is very practical. The `--manifest` parameter should support it too.

spyrkob:
Closes #726


spyrkob:
@TomasHofman FYI

TomasHofman:
Looks good! Is there any handling necessary in the `channel add` command, as you pointed out in the issue? Throwing error if there is more than one manifest defined..?

spyrkob:
`channel add` operation should already only accept a single manifest (defined through the `@CommandLine.Option` annotations), that should not be affected by this changes, unless I missed something :)

TomasHofman:
I see, it doesn't inherit from the AbstractInstallCommand. Sounds good :).

TomasHofman (approved):

spyrkob:
Thanks for the review @TomasHofman :)

Issue #657, opened 2024-05-09
Print more helpful message if no artifact matching the version pattern is found


xstefank:
Current behavior:

```
ERROR: Unable to resolve artifacts:
  * io.smallrye.config:smallrye-config-common:jar [missing]
 Attempted repositories:
  *MRRC::https://maven.repository.redhat.com/ga/
  *indy::https://indy.psi.redhat.com/api/content/maven/hosted/pnc-builds/
  *brew::https://download.devel.redhat.com/brewroot/repos/jb-eap-8.0-maven-build/latest/maven/
  *central::https://repo1.maven.org/maven2/
```

My issue was that I had `versionPattern: ".*-redhat-.*"` and the artifacts are `.redhat-`. Would be good to have more details in the error message.

ChristinaDsl:
@spyrkob I will take a look at it

parsharma:
@ChristinaDsl are you working on this issue?

ChristinaDsl:
@parsharma for now no, you can take it if you wish :)

PR #738, opened 2024-07-25
Print more helpful message if no artifact matching the version pattern is found

parsharma:
Issue: https://github.com/wildfly-extras/prospero/issues/657

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
@@ -176,9 +176,21 @@ public class ExecutionExceptionHandler implements CommandLine.IExecutionExceptio
             }
 
             printRepositories(ex.getRepositories(), ex.isOffline());
+            if (versionPatternMismatch(ex.getMissingArtifacts())) {
+                console.error("It seems the version pattern does not match the expected format. Ensure that your version pattern is correct.");
+            }
         } else {
             console.error(CliMessages.MESSAGES.errorHeader(ex.getLocalizedMessage()));
         }
+    }
+    private boolean versionPatternMismatch(Set<ArtifactCoordinate> missingArtifacts) {
+        for (ArtifactCoordinate artifact : missingArtifacts) {
+            // Adjust this pattern to match your specific needs
+            if (!artifact.getVersion().matches(".*\\.redhat-.*")) {
+                return true;
+            }
+        }
+        return false;
     }
 
     private void printRepositories(Set<Repository> repositories, boolean b) {


spyrkob (changes requested):
@parsharma this has to work for any patterns not just this specific one text. It probably needs to be detected at wildfly-channel library level and reported back in the exception.

TomasHofman:
I was looking with @parsharma , I think if we want to improve the message to give more details, we would have to essentially list details about all involved channels, so the original message:

```
ERROR: Unable to resolve artifacts:
  * io.smallrye.config:smallrye-config-common:jar [missing]
  * [other artifacts...]
 Attempted repositories:
  *MRRC::https://maven.repository.redhat.com/ga/
  *indy::https://indy.psi.redhat.com/api/content/maven/hosted/pnc-builds/
  *brew::https://download.devel.redhat.com/brewroot/repos/jb-eap-8.0-maven-build/latest/maven/
  *central::https://repo1.maven.org/maven2/
```

would probably need to look like:

```
ERROR: Unable to resolve artifacts:
  * io.smallrye.config:smallrye-config-common:jar [missing]
    [here search the channel streams if they contains given artifact and show the version or version pattern, like:]
    * channel A: stream found with version pattern "blah"
    * channel B: stream found with version "1.2.3"
    * channel C: no stream found, no stream strategy: LATEST
  * ...
 Attempted repositories:
  [here show repositories defined in each channel:]
  * channel A:
    *MRRC::https://maven.repository.redhat.com/ga/
    *indy::https://indy.psi.redhat.com/api/content/maven/hosted/pnc-builds/
    *brew::https://download.devel.redhat.com/brewroot/repos/jb-eap-8.0-maven-build/latest/maven/
  * channel B:
    *central::https://repo1.maven.org/maven2/
  ...
```

It could get quite verbose in the end, would have to somehow display details about all channels relevant to each missing artifact.

Maybe necessarily we wouldn't need to modify the channel library to provide more info in the original exception, as we can get access to the list of channels ourselves in prospero, we know that all configured channels are used by the channel session.

The effort and verbosity in the end doesn't look awful, but still it's to be considered if we should go into these details, to me it looks helpful for end user unless it gets too verbose.

WDYT?

TomasHofman:
Then also would that be like giving too much information to end user? Maybe it's useful to person assembling the channels, but too much info for the person running the installation..? Maybe print that in some verbose mode? Just ideas...

spyrkob:
Thanks for the analysis @TomasHofman!

Keeping it in verbose mode might make sense. It should be a fairly small change to switch it between verbose and normal output, so we can see how it feels when it's implemented. 

I was originally thinking more along the lines of a short message like  ` * io.smallrye.config:smallrye-config-common [missing - no version matching patterns "<PATTERN,...>"]` if the only channels providing this stream use a versionPattern. But maybe that's a good opportunity to review other errors and provide better logging. I'll take a closer look at this next week.

Issue #759, opened 2024-09-11
Update process should check version and not allow reversion through update


parsharma:
Our partners in Japan have discovered that it is possible to "revert" an update by "updating" to the previous version by following the steps in [3.2. Updating JBoss EAP offline using the jboss-eap-installation-manager](https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/8.0/html-single/updating_red_hat_jboss_enterprise_application_platform/index#proc_updating-jboss-eap-8-0-offline-using-the-installer-manager_default) and using an earlier version's repository; i.e. one can go from 8.0.1.1 to 8.0.1 by doing an "update perform" with the 8.0.1 maven-repository. The update will be shown as an "update" in the history, even though technically it was a downgrade/reversion.

$ /opt/rh/eap8.0/jboss-eap-8.0/bin/jboss-cli.sh -c "installer history"
[22bd37d2] 2024-05-09T04:57:07Z - update [org.jboss.eap.channels:eap-8.0::1.1.0.GA-redhat-00001]   <-- update (downgrade?) to 8.0.1
[ca214488] 2024-05-09T04:56:07Z - update [org.jboss.eap.channels:eap-8.0::1.1.1.GA-redhat-00002]   <-- update to 8.0.1.1
[3268b557] 2024-01-19T09:04:16Z - install [org.jboss.eap.channels:eap-8.0::1.0.1.GA-redhat-00003]  <-- install 8.0.1

The update process should check the current and "new" version of the patch (1.1.1.GA-redhat-00002 vs 1.1.0.GA-redhat-00001) and not allow the user to install an older version, perhaps informing them to use "revert" instead.Our partners in Japan have discovered that it is possible to "revert" an update by "updating" to the previous version by following the steps in [3.2. Updating JBoss EAP offline using the jboss-eap-installation-manager](https://access.redhat.com/documentation/en-us/red_hat_jboss_enterprise_application_platform/8.0/html-single/updating_red_hat_jboss_enterprise_application_platform/index#proc_updating-jboss-eap-8-0-offline-using-the-installer-manager_default) and using an earlier version's repository; i.e. one can go from 8.0.1.1 to 8.0.1 by doing an "update perform" with the 8.0.1 maven-repository. The update will be shown as an "update" in the history, even though technically it was a downgrade/reversion.

$ /opt/rh/eap8.0/jboss-eap-8.0/bin/jboss-cli.sh -c "installer history"
[...]

PR #765, opened 2024-09-23
Update process should check version and not allow reversion through update

parsharma:
Issue: https://github.com/wildfly-extras/prospero/issues/759

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
@@ -29,6 +29,7 @@ import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 import org.apache.commons.io.FileUtils;
+import org.eclipse.aether.artifact.Artifact;
 import org.jboss.galleon.ProvisioningException;
 import org.jboss.galleon.diff.FsDiff;
 import org.jboss.galleon.diff.FsEntry;
@@ -39,6 +40,7 @@ import org.wildfly.channel.Channel;
 import org.wildfly.channel.ChannelManifest;
 import org.wildfly.channel.ChannelManifestCoordinate;
 import org.wildfly.channel.Repository;
+import org.wildfly.channel.version.VersionMatcher;
 import org.wildfly.prospero.ProsperoLogger;
 import org.wildfly.prospero.actions.ApplyCandidateAction;
 import org.wildfly.prospero.actions.SubscribeNewServerAction;
@@ -60,6 +62,7 @@ import org.wildfly.prospero.cli.ReturnCodes;
 import org.wildfly.prospero.api.TemporaryFilesManager;
 import org.wildfly.prospero.galleon.FeaturePackLocationParser;
 import org.wildfly.prospero.galleon.GalleonUtils;
+import org.wildfly.prospero.metadata.ManifestVersionRecord;
 import org.wildfly.prospero.metadata.ProsperoMetadataUtils;
 import org.wildfly.prospero.model.InstallationProfile;
 import org.wildfly.prospero.updates.UpdateSet;
@@ -120,7 +123,6 @@ public class UpdateCommand extends AbstractParentCommand {
                 log.tracef("Perform full update");
 
                 console.println(CliMessages.MESSAGES.updateHeader(installationDir));
-
                 try (UpdateAction updateAction = actionFactory.update(installationDir, mavenOptions, console, repositories)) {
                     performUpdate(updateAction, yes, console, installationDir, noConflictsOnly);
                 }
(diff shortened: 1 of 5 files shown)


TomasHofman:
Relevant wildfly-channel issue that I think we would need implemented there in order to complete this functionality:
https://github.com/wildfly-extras/wildfly-channel/issues/292

TomasHofman:
A good next step could be to obtain a list of manifests used by the current installation. I think it should be possible to obtain maven GAVs (or URLs) of the current manifests. If there are any Maven based manifest found in the list, they should be matched against the list of new manifests (which the user want to upgrade to) used by the ChannelSession.

(That list of new manifests is currently simulated by the `findManifestUpdates()` method, hard-coded to only return `org.jbosss.eap.channels:eap-8.0` manifest artifact, because we don't know how to obtain these data.)

parsharma:
Hello @TomasHofman,
I have updated the code where we are getting list of manifests from the current installation.

TomasHofman:
Now can we get the list of manifests currently used by the installation? Maybe from the InstallationMetadata class?

If we can have that, we can try to obtain an intersection from those two lists (manifests groupId:artifactId s that are present in both currently installed manifest and the newly available manifests used by the channelSession).

Once we have that intersection, we can try to compare the versions to see if they are upgrading or downgrading.

parsharma:
@TomasHofman 
I have added the new method compareManifest() where this method compares the currently installed Maven manifests with the available manifest updates from the ChannelSession. It finds the manifests that exist in both the installed manifests and the available updates (based on groupId and artifactId).  For each matching manifest, it compares the version of the installed manifest with the version from the available update. Depending on the comparison, the method outputs whether an upgrade, downgrade, or no change is required.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
```diff
@@ -471,4 +478,49 @@ public static Path detectProsperoInstallationPath() throws ArgumentParsingExcept
...
+
+            if (updateArtifact != null) {
+                // Compare versions
+                String installedVersion = installedManifest.getVersion();
+                String availableVersion = updateArtifact.getVersion();
+
+                if (installedVersion.compareTo(availableVersion) < 0) {
+                    System.out.println("Upgrade available for " + installedManifest.getArtifactId() + ": " + installedVersion + " -> " + availableVersion);
+                } else if (installedVersion.compareTo(availableVersion) > 0) {
+                    System.out.println("Downgrade detected for " + installedManifest.getArtifactId() + ": " + installedVersion + " -> " + availableVersion);
```
TomasHofman, line 518: I think the comparison logic here should not be an alphabetic comparison, but something that understands semantic versioning.

The problem is that e.g. version "1.11" would be evaluated as lower than "1.2" when using alphabetical ordering, while it should  be higher semantically.

parsharma: Thanks for noticing it @TomasHofman 
I have the code where I am using DefaultArtifactVersion which we compare the versions semantically.

spyrkob: `org.wildfly.channel.version.VersionMatcher` has the correct version compare logic

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
```diff
@@ -471,4 +479,49 @@ public static Path detectProsperoInstallationPath() throws ArgumentParsingExcept
...
+                if (update.getGroupId().equals(installedManifest.getGroupId()) &&
+                        update.getArtifactId().equals(installedManifest.getArtifactId())) {
+                    intersectingManifests.add(installedManifest); // Add to intersection list
+                    break;
+                }
+            }
+        }
+
+        // For each intersecting manifest, compare the version
+        for (ManifestVersionRecord.MavenManifest installedManifest : intersectingManifests) {
```
spyrkob, line 499: you don't need the `intersectingManifest` list. Just iterate over mavenManifests. You're already ignoring ones that are not in `updateArtifacts`.

parsharma:
The CI/CD is failing because we need wildfly-channel upgrade for the API.

TomasHofman:
Yes, we must ask Jeff to release.

TomasHofman:
Never mind, he did and you upgraded already.

TomasHofman:
@spyrkob currently this only checks for downgrade in the `update perform` command. There is the second pathway with `update prepare` / `update apply`. Should the check be made during `prepare` or `apply` stage? If I prepare and update, and then the installation changes, is the prepared update still applicable or do I need to prepare it again?

TomasHofman:
Maybe we could move the check into `UpdateCommand#buildUpdate()` method, which would perform the check both during `update perform` and `update prepare`.

spyrkob:
>  I prepare and update, and then the installation changes, is the prepared update still applicable or do I need to prepare it again?

If the base server changes from the state that the candidate was generated from, that candidate is no longer valid and cannot be applied.

Issue #755, opened 2024-09-06
`channel list` output is confusing


jmesnil:
The output of `channel list` is confusing... It looks like YAML file but it's not. It looks like a [channel definition](https://github.com/wildfly-extras/wildfly-channel/blob/main/doc/spec.adoc#channel-definition) but it's not.

```
$ prospero channel list
Server /Users/jmesnil/Developer/wildfly/dist/target/wildfly-34.0.0.Beta1-SNAPSHOT/. is subscribed to following channels:

-------
# wildfly
  manifest: org.wildfly.channels:wildfly:34.0.0.Beta1-SNAPSHOT
  repositories:
    id: jboss-public-repository-group
    url: https://repository.jboss.org/nexus/content/groups/public/
    id: central
    url: https://repo.maven.apache.org/maven2
    id: jboss-ga-repository
    url: https://maven.repository.redhat.com/ga/
-------
```

* `# wildfly` actually correspond to the `name` of the channel.
* `manifest` is a different representation of the manifest coordinates (Maven GAV or URL)
* `repositories` kind of represents array items

I'm not sure to understand why this command does not output the *actual* content of the `.installation/installer_channels.yaml`:

```
$ cat .installation/installer-channels.yaml
---
schemaVersion: "2.0.0"
name: "wildfly"
repositories:
  - id: "jboss-public-repository-group"
    url: "https://repository.jboss.org/nexus/content/groups/public/"
  - id: "central"
    url: "https://repo.maven.apache.org/maven2"
  - id: "jboss-ga-repository"
    url: "https://maven.repository.redhat.com/ga/"
manifest:
  maven:
    groupId: "org.wildfly.channels"
    artifactId: "wildfly"
    version: "34.0.0.Beta1-SNAPSHOT"
noStreamStrategy: "none"
```

Maybe we could provide a short and long output for `channel list`.
The short list would display the name of the channel as well as the manifest GAV or URL (which are the most important information) while the long list would dump `.installation/installer_channels.yaml` (which would be a valid YAML file corresponding to channel definitions).

Something like:

```
$ prospero channel list
wildfly org.wildfly.channels:wildfly:34.0.0.Beta1-SNAPSHOT

$ prospero channel list --full
---
schemaVersion: "2.0.0"
name: "wildfly"
repositories:
  - id: "jboss-public-repository-group"
    url: "https://repository.jboss.org/nexus/content/groups/public/"
  - id: "central"
    url: "https://repo.maven.apache.org/maven2"
[...]

PR #777, opened 2024-10-16
channel list output is confusing #755

parsharma:
Issue: https://github.com/wildfly-extras/prospero/issues/755

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/ChannelCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/ChannelCommand.java
@@ -20,14 +20,17 @@ package org.wildfly.prospero.cli.commands;
 import java.nio.file.Path;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import org.wildfly.channel.Channel;
+import org.wildfly.channel.ChannelManifestCoordinate;
+import org.wildfly.channel.ChannelMapper;
+import org.wildfly.channel.Repository;
 import org.wildfly.prospero.actions.MetadataAction;
 import org.wildfly.prospero.cli.ActionFactory;
 import org.wildfly.prospero.cli.CliConsole;
 import org.wildfly.prospero.cli.CliMessages;
 import org.wildfly.prospero.cli.ReturnCodes;
-import org.wildfly.prospero.cli.printers.ChannelPrinter;
 import org.wildfly.prospero.metadata.ManifestVersionRecord;
 import picocli.CommandLine;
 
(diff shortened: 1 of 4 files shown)


prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/ChannelCommand.java
```diff
@@ -53,6 +53,9 @@ public static class ChannelListCommand extends AbstractCommand {
         @CommandLine.Option(names = CliConstants.DIR)
         Optional<Path> directory;
 
+        @CommandLine.Option(names = CliConstants.FULL)
```
spyrkob, line 56: New options need to be described in the `UsageMessages.properties`

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/ChannelCommandTest.java
```diff
@@ -191,7 +191,33 @@ public void testList() {
         int exitCode = commandLine.execute(CliConstants.Commands.CHANNEL, CliConstants.Commands.LIST,
                 CliConstants.DIR, dir.toString());
         Assert.assertEquals(ReturnCodes.SUCCESS, exitCode);
-        Assert.assertEquals(3, getStandardOutput().lines().filter(l->l.contains("manifest")).count());
+        Assert.assertEquals(3, getStandardOutput().lines()
```
spyrkob, line 194: Use `Assertions.assertThat` and compare it to actual expected lines ("test1 g:a", "test2 g:a:v", "test3 file:/a:b")

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/ChannelCommandTest.java
```diff
@@ -191,7 +191,33 @@ public void testList() {
...
+                .count());
+    }
+
+    @Test
+    public void testFullList() {
+        int exitCode = commandLine.execute(CliConstants.Commands.CHANNEL, CliConstants.Commands.LIST, CliConstants.FULL,
+                CliConstants.DIR, dir.toString());
+        Assert.assertEquals(ReturnCodes.SUCCESS, exitCode);
+        String output = getStandardOutput();
+        Assert.assertTrue("Expected 'schemaVersion' in the output",
```
spyrkob, line 205: The test doesn't actually verify that the correct values are printed.  You should be able to get the output, convert it back to Channel objects and compare with the input. 
Or simply compare the output with serialized input.

spyrkob (changes requested):

parsharma:
@spyrkob, I have updated the code according to the review comments. Please have a look and let me know if more changes are needed.
Thanks

PR #813, opened 2024-12-06
Downgrade plexus-utils from 3.5.0 to 3.0.24

TomasHofman:
This is to align with wildfly-jar-maven-plugin.


spyrkob:
Thanks @TomasHofman!

PR #837, opened 2025-02-12
[#836] When applying a candidate only files managed by the Galleon should be backed-up and rolled back

### LINKED ISSUE #836: When applying a candidate only files managed by the Galleon should be backed-up and rolled back
When a file or a folder owned by another user is present in the installation folder and the user account running update has read-only permission to the folder, prospero will fail when generating backup snapshot.

Such folders should be ignored and not managed by the backup - either when creating snapshot or restoring from it.

spyrkob:
Fixes #836


integration-tests/src/test/java/org/wildfly/prospero/extended/MultiUserServerDirectoryTest.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Test
+    public void updateServerWithAnExternalFolderNotOwnedByUser() throws Exception {
+        // actual test setup - make the server and prospero owned by 'serveradmin'
+        container.execInContainer("chmod", "0777", "serveradmin:serveradmin", "/home/serveradmin");
+        container.execInContainer("chmod", "0777", "serveradmin:serveradmin", "/home/serveradmin/wildfly");
+        container.execInContainer("chown", "-R", "serveradmin:serveradmin", "/home/serveradmin");
+
+        // actual test setup - create the file owned be a different user
```
TomasHofman, line 143: s/be/by

integration-tests/src/test/java/org/wildfly/prospero/extended/MultiUserServerDirectoryTest.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    @Test
+    public void updateRollbackDoesNotRemoveExternalFiles() throws Exception {
+        // actual test setup - make the server and prospero owned by 'serveradmin'
+        container.execInContainer("chmod", "0777", "serveradmin:serveradmin", "/home/serveradmin");
+        container.execInContainer("chmod", "0777", "serveradmin:serveradmin", "/home/serveradmin/wildfly");
+        container.execInContainer("chown", "-R", "serveradmin:serveradmin", "/home/serveradmin");
+
+        // actual test setup - create the file owned be a different user
```
TomasHofman, line 120: s/be/by

integration-tests/src/test/java/org/wildfly/prospero/extended/MultiUserServerDirectoryTest.java
```diff
@@ -0,0 +1,239 @@
...
+        final Container.ExecResult updateRes = serveradmin.tryExecute("/home/serveradmin/prospero/bin/prospero.sh", "update", "perform",
+                "--dir", "/home/serveradmin/wildfly", "--repositories", "/home/serveradmin/repository/", "-y", "-vv");
+        assertEquals(updateRes.toString(), 1, updateRes.getExitCode());
+
+        final Container.ExecResult res = container.execInContainer("ls", "-alF", "/home/serveradmin/wildfly");
+        assertTrue(res.toString(), filesystem.fileExists("/home/serveradmin/wildfly/test"));
+    }
+
+    @Test
+    public void updateServerWithAnExternalFolderNotOwnedByUser() throws Exception {
```
TomasHofman, line 137: What is the "external folder"? The "/home/serveradmin/wildfly/test" looks to be a file.

spyrkob: Good point, I'll add a folder owned by the siteadmin to cover that case.

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyStageBackup.java
```diff
@@ -65,55 +67,106 @@ class ApplyStageBackup implements AutoCloseable {
...
 
-            @Override
-            public FileVisitResult visitFileFailed(Path file, IOException exc) throws IOException {
-                // if the file in the installation is not readable AND it does not exist in the candidate
-                // we assume it is user controlled file and we ignore it
-                return ignoreUserManagedFiles(file, exc);
-            }
+        // if the server doesn't conatain a galleon provisioning record, skip building the backup
+        if (!Files.exists(hashesRoot)) {
+            ProsperoLogger.ROOT_LOGGER.warn("Unable to perform the backup: No Galleon Hashes record found.");
```
TomasHofman, line 80: Are we at risk that this version of prospero will by used with an instance provisioned by an old prospero instance? In which case it would become unusable?

prospero-common/src/main/java/org/wildfly/prospero/actions/GalleonHashesFileVisitor.java
```diff
@@ -0,0 +1,62 @@
...
+            // read hashes descriptor
+            final List<String> readLines = FileUtils.readLines(file.toFile());
+
+            // read every other line - we don't care about hashes atm.
+            for (int i = 0; i < readLines.size(); i+=2) {
+                final String recordedFileName = readLines.get(i);
+
+                doVisitFile(currentDirectory.resolve(recordedFileName));
+            }
+        }
```
TomasHofman, line 45: Is this guaranteed that files will always be listed in the expected order - and we process only even lines?

prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyStageBackup.java
spyrkob: Do you mean if the galleon `.provisioned/hashes` folder is not present? That folder should be present in any instance provisioned by Prospero. Or is there some other change that might be problematic?

If the hashes folder doesn't exist, the update should still proceed, just without the backup. Which actually I should add to the testsuite :)

TomasHofman: Yeah, I'm missing that context, I thought the hashes file is a new thing. This looks OK then :).

TomasHofman (commented):
I'm still bit grappling with the context :).

To sum it up, it sorta looks like we only take notice of the files listed in the hashes files, that are created by galleon during provisioning. I assume hashes contain all files generated by provisioning.

In practice, when I provision an installation, and then manually add a e.g. a custom module directory, that directory would be left out of this processing. Does it mean the directory will be present or missing after installation update?

spyrkob:
@TomasHofman this doesn't change the current update behaviour. The user created/modifed files should not be modified or removed *unless* they are marked as system files by the feature pack.


The goal of the backup is to be able to revert update changes if during `apply` state prospero runs into an issue (e.g. unable to replace updated file). In order not to finish the update in an invalid state, we use the backup to restore the files.

Originally I tried to backup all the files in the server directory, but that run into an issue when some of the added files were not read/write -able by the user running prospero update.

As a fix the backup now takes into account only files that are managed by the feature pack - i.e. ones recorded in the Galleon `.hashes` folder. The thinking is that any other files should not be touched by the `update apply` process so they do not need to be updated.

Let me know if that makes it a bit more clear

TomasHofman:
OK, so if the non-managed files remain in the installation after update/rollback (I'm not sure about these details), it looks all right to me :). And it makes sense to exclude them from the backup process.

prospero-common/src/main/java/org/wildfly/prospero/actions/GalleonHashesFileVisitor.java
spyrkob: @TomasHofman I replaced the parsing/tree walking with Galleon API here

spyrkob:
Thanks for the review @TomasHofman !

PR #916, opened 2025-07-22
JBEAP-29588 Fix InstallCommand to handle zipped repositories

TomasHofman:
Issue: https://issues.redhat.com/browse/JBEAP-29588
Upstream issue: N/A
Upstream PR: https://github.com/wildfly-extras/prospero/pull/917

* In the install command, the unpacking functionality is only applied on the shadowRepositories field, but not on the remoteRepositories field.
* Rename TemporaryFilesManager.getInstance() to TemporaryFilesManager.newInstance(), because the method really always creates a new instance, and doesn't handle an existing instance in any way.


TomasHofman:
Manual test was done using command like:

```
./prospero install --definition=/home/thofman/prospero/eap-8.1-provisioning.xml --manifest org.jboss.eap.channels:eap-8.1 --repositories=/home/thofman/prospero/jboss-eap-8.1.0.GA-CR9-runtime-maven-repository.zip,/home/thofman/prospero/jboss-eap-8.1.0.GA-CR9-runtime-maven-repository.zip --dir=eap-8.1again --offline
```

as upstream prospero doesn't know about EAP profiles, so one has to use the --definition and --channel parameters instead of --profile.

pedro-hos:
@TomasHofman one think that I just notice now, this PR title and commit message should point to JBEAP-29588, related to 1.3 version (EAP 8.1) the JBEAP-29587 1.1 version is OK -> https://github.com/wildfly-extras/prospero/pull/918 

Thank you

TomasHofman:
Yeah I know. I was thinking it's bit stupid to keep changing Jira codes when porting commits up and down, as they are clone of the same issue anyway.

Usually in other components we would mostly have just the component issue code in the commit (like ELY-...), not the JBEAP codes, but prospero doesn't have a Jira project and I don't think it needs one.

(I was thinking about removing issue codes altogether, from the commit / PR descriptions, but then I think it's useful to have some reference as to what triggered the change.)

I will fix it. :)

PR #925, opened 2025-07-29
[JBEAP-30585] - JBoss EAP 8 Update Deletes 'log' Directory Symlink, Causing Startup Failure

pedro-hos:
Issue: [https://issues.redhat.com/browse/JBEAP-30585](https://issues.redhat.com/browse/JBEAP-30585)

I have made a validation for Symlink during the delete time. So, If the directory is a Symlink, we didn't remove. Also added some tests.

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyCandidateAction.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/actions/ApplyCandidateAction.java
@@ -681,6 +681,7 @@ public class ApplyCandidateAction {
 
         // Delete the files in the installation that are not present in the update and not added by the user
         // We need to skip .glnew and .glold.
+        // JBEAP-30585: We need to keep the Symbolic files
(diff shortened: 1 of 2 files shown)


TomasHofman:
I noticed Yeray reviewed the other PR, will read up.

jamezp:
Do we need to keep symbolic files or symbolic directories? Potentially both, but the directory seems important here. IoW it seems like we should be clean the log directory as we normally would, but leaving the symbolic link to the directory itself alone.

TomasHofman:
@jamezp you may want to check the discussion on the downstream PR (https://github.com/wildfly-extras/prospero/pull/926), where Yeray had some input.

> we should be clean the log directory as we normally would...

Regarding that, both me and Yeray checked and it looks that we normally *do not* clean the log directory. Essentially any files that are not created during provisioning are taken as user-made changes, and are retained in the original installation. That also applies to log files as those are as well created outside of provisioning process, and there is no special handling around them.

PR #944, opened 2025-08-29
Update URL

jmesnil:
after the project was transferred to the wildfly organization on GitHub


TomasHofman (approved):
We should as well port this into the 1.3.x, where we very likely will do additional releases.

When porting let's do full-text search for "wildfly-extras" to double check that there are no remaining strings on that branch referring to extras.

PR #958, opened 2025-09-16
JBEAP-30876 - (8.0.8) installation-manager >8.0.4 update fails with symbolic link

pedro-hos:
Upstream: NA
Jira: https://issues.redhat.com/browse/JBEAP-30876

It's not possible to create hardlinks on a different filesystem and it throws `FileSystemException` instead of `UnsupportedOperationException`:

```
Caused by: java.nio.file.FileSystemException: usb/.update.old/domain/configuration/mgmt-users.properties -> usb/domain/configuration/mgmt-users.properties: Invalid cross-device link
```


TomasHofman (approved):
Good job!

pedro-hos:
@TomasHofman  good point, I'll create the test case. Yes, we'll need to porting it into `main` and `1.3` branches: https://issues.redhat.com/browse/JBEAP-30951

PR #964, opened 2025-09-23
[Issue_963] Add javadoc and sources jars for the dist/build module

### LINKED ISSUE #963: org.wildfly.prospero:prospero-build must produce source and javadoc jars
https://issues.redhat.com/browse/NEXUS-1046

Validation of the org.wildfly.prospero:prospero-build 1.1.9.Final artifacts fail validation for syncing to Maven Central because it does not produce -javadoc.jar or -sources.jar artifacts. Even though such artifacts may seem non-sensical, particularly javadoc, they must be produced.

In other cases similar to these we have updated the relevant pom to produce and attach empty artifacts.

bstansberry:
Resolves #963

### DIFF
--- a/dist/build/pom.xml
+++ b/dist/build/pom.xml
@@ -115,6 +115,38 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <executions>
+                    <execution>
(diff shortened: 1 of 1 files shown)


TomasHofman:
Ah, this escaped me, I should have (and planned) porting that commit to the remaining branches. Thanks @bstansberry!

@pedro-hos we can close this PR and merge #966 and #965 if it looks all right to you.

PR #966, opened 2025-09-23
Change the dist/build/ submodule to pom packaging, to avoid producing…

### LINKED ISSUE #963: org.wildfly.prospero:prospero-build must produce source and javadoc jars

bstansberry:
… unneeded jar

This break JBoss Nexus and possibly Maven Central validations.

Resolves #963 in the 1.1.x branch.

Upstream: #965

The 1.3.x branch already has this via https://github.com/wildfly/prospero/commit/b9ac334d2b3dba5441850fd5b360f43b18f7b1dd

### DIFF
--- a/dist/build/pom.xml
+++ b/dist/build/pom.xml
@@ -7,7 +7,7 @@
 
     <modelVersion>4.0.0</modelVersion>
     <artifactId>prospero-build</artifactId>
-    <packaging>jar</packaging>
+    <packaging>pom</packaging>
 
     <name>Prospero Build</name>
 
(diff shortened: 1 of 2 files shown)


pedro-hos:
Thank you @bstansberry

PR #981, opened 2025-11-06
[ISSUE-980] Zip with no folder entries working in prospero test added.

elguardian:
Closes: https://github.com/wildfly/prospero/issues/980

### DIFF
--- a/prospero-common/src/test/java/org/wildfly/prospero/api/RepositoryUtilsTest.java
+++ b/prospero-common/src/test/java/org/wildfly/prospero/api/RepositoryUtilsTest.java
@@ -74,6 +74,7 @@ public class RepositoryUtilsTest {
         this.zipHandler = zipHandler;
     }
 
+    @Test
     public void unzipArchiveBeforeUsingIt() throws Exception {
         final Path repoRoot = temp.newFolder("repo").toPath();
         final Path zipFile = createRepository(repoRoot);


prospero-common/src/test/java/org/wildfly/prospero/api/RepositoryUtilsTest.java
```diff
@@ -17,34 +17,63 @@
...
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.junit.Assert.assertEquals;
 
+@RunWith(Parameterized.class)
 public class RepositoryUtilsTest {
 
     @Rule
     public TemporaryFolder temp = new TemporaryFolder();
 
-    @Test
```
TomasHofman: Was it intentional to remove the `@Test` annotation?

pedro-hos: I guess that we don't need the `@Test` with `@Parameters`. We can use only the `@Parameters` here. Is this  the reason for you remove the `@Test` @elguardian?

elguardian: good catch. It was not intentional. Anyway this just verify the piece of code that handles repo.

elguardian:
closing this as I did check zip file system is working either with folder entries or not. There is nothing else in the jira with any specifics about any further problems.

TomasHofman:
:+1: , thanks!

PR #1010, opened 2026-02-25
JBEAP-32052 - jboss eap installation manager (prospero) update perform returns exit code 1 when no updates found instead of 0

pedro-hos:
issue: https://issues.redhat.com/browse/JBEAP-32052


TomasHofman:
Hi Pedro, looks good to me. I think we could add a test case for this to guard it doesn't break in future, and it would make it easier for QE to verify.

PR #1013, opened 2026-03-06
When --debug is enabled, the embedded process shouldn't be forked

### LINKED ISSUE #1011: When --debug is enabled, the embedded process shouldn't be forked
To get all traces (we are missing the ones that are printed by JBoss Logging by the embedded server), the embedded server must run inside the Prospero process and benefit from the JBoss Log Manager configuration.
The line that should be conditioned by the debug mode: https://github.com/wildfly/prospero/blob/main/prospero-common/src/main/java/org/wildfly/prospero/galleon/GalleonUtils.java#L82

pedro-hos:
Closes: #1011

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/galleon/GalleonUtils.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/galleon/GalleonUtils.java
@@ -17,16 +17,6 @@
 
 package org.wildfly.prospero.galleon;
 
-import org.apache.commons.io.FileUtils;
-import org.jboss.galleon.ProvisioningException;
-import org.jboss.galleon.diff.FsDiff;
-import org.jboss.galleon.universe.FeaturePackLocation;
-import org.jboss.galleon.universe.UniverseResolver;
-import org.jboss.galleon.universe.maven.repo.MavenRepoManager;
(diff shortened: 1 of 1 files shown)


TomasHofman:
Will defer to @jfdenise , I don't see a problem.

IIUC, in production use Prospero always runs in modular environment (either as standalone provisioned feature pack, or as feature pack provision as part of Wildfly/EAP), so this would only affect the "working" build in "dist/build/" module, when building prospero from sources..?

TomasHofman:
Will defer to @jfdenise , I don't see a problem.

IIUC, in production use Prospero always runs in modular environment (either as standalone provisioned feature pack, or as feature pack provision as part of Wildfly/EAP), so this would only affect the "working" build in "dist/build/" module, when building prospero from sources..?

pedro-hos:
Yes, this will only works in a non modular way, when we run the `./prospero` at the root path from the Prospero project, that's calls the `prospero-cli` jar directly.

pedro-hos:
@jfdenise do you have some inputs for this issue?

jfdenise:
Actually I think that we can close this PR, the use case doesn't seem to be worth the changes.

PR #1016, opened 2026-03-11
[JBEAP-32046] Fix MissingFormatArgumentException in ExecutionExceptio…

rafaeldarosagit:
…nHandler

When a repository URL contains %20, String.format() fails. This commit escapes percent characters and adds unit tests.

### DIFF
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
@@ -201,7 +201,9 @@ public class ExecutionExceptionHandler implements CommandLine.IExecutionExceptio
             for (Repository repository : repositories) {
                 boolean isOffline = isOffline(b, repository);
                 String repo = String.format("%s::%s", repository.getId(), repository.getUrl());
-                console.error("  *" + repo + (isOffline ? " ["+ CliMessages.MESSAGES.offline() + "]" : ""));
+                // Escape percent characters to prevent String.format() from interpreting them as format specifiers
+                String escapedRepo = repo.replace("%", "%%");
+                console.error("  *" + escapedRepo + (isOffline ? " ["+ CliMessages.MESSAGES.offline() + "]" : ""));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


rafaeldarosagit:
Link issue: https://issues.redhat.com/browse/JBEAP-32046

prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
```diff
@@ -201,7 +201,9 @@ private void printRepositories(Set<Repository> repositories, boolean b) {
             for (Repository repository : repositories) {
                 boolean isOffline = isOffline(b, repository);
                 String repo = String.format("%s::%s", repository.getId(), repository.getUrl());
```
pedro-hos: LGTM, just a small thing, but instead creating another variable, you can scape the "%" directly here, like:

`String repo = String.format("%s::%s", repository.getId(), repository.getUrl()).repllace("%", "%%");`. 

WDYT?

prospero-cli/src/test/java/org/wildfly/prospero/cli/ExecutionExceptionHandlerTest.java
```diff
@@ -0,0 +1,168 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+
+public class ExecutionExceptionHandlerTest {
```
pedro-hos, line 38: Great work here Rafa! Thank you

pedro-hos (approved):
Thank you

prospero-cli/src/main/java/org/wildfly/prospero/cli/ExecutionExceptionHandler.java
rafaeldarosagit: Good catch! I've updated the code to escape the '%' directly without the extra variable. Thank you!

PR #1024, opened 2026-03-13
JBEAP-32344 - NPE in installer update when a new channel was added to the installation

pedro-hos:
Issue: https://issues.redhat.com/browse/JBEAP-32344


pedro-hos:
@TomasHofman do you think that we need to change someting at the Instalation Manager API level? Now, we have the `oldChannel` null, same for `location`.

pedro-hos:
suppressed by https://github.com/wildfly/prospero/pull/1026

PR #1026, opened 2026-03-16
JBEAP-32344 NPE in installer update when a new channel was added to t…

TomasHofman:
…he installation

Issue: https://issues.redhat.com/browse/JBEAP-32344
Upstream not required

Fixes NPE when listing possible upgrades in a scenario when a new channel has been added before the upgrade.


pedro-hos (approved):
Thank you @TomasHofman

PR #1028, opened 2026-03-19
Fixing NPEs during upgrade operation

TomasHofman:
These happen in a scenario when upgrade is called after new channel has been subscribed, which was not part of the installation originally.


pedro-hos (approved):
Thank you Tomas.

PR #1036, opened 2026-04-29
Csd modifications

TomasHofman:
This introduces:

* The `--maven-settings` parameter to point prospero to the maven settings.xml file (defaults to ~/.m2/settings.xml). The settings file is then used by prospero to extract authentication credentials for Maven repositories.
* The `--filter-manifest`, that adds the ability to add an additional method of filtering available artifacts. This is used in case that the main input channel uses versionPatterns or no versions at all and resolves to the latest available artifact version. We would presently use it to filter out dependencies for specific EAP release (8.1 / 8.2) from a shared maven repo.
* Updates the wildfly-channel lib and prospero-metadata to versions that support accessing authenticated files over HTTP. This is used in cases when some channel metadata (manifest, blocklist) are referred as URLs rather, than Maven coordinates, and require HTTP authentication. The auth token can be provided as env variable ()`WILDFLY_CHANNEL_HTTP_AUTH_TOKEN` or system property (`org.wildfly.channel.http.auth.token`). See https://github.com/wildfly/wildfly-channel/pull/402

### DIFF
--- a/dist/common/src/main/resources/modules/system/layers/base/org/jboss/prospero/main/module.xml
+++ b/dist/common/src/main/resources/modules/system/layers/base/org/jboss/prospero/main/module.xml
@@ -46,11 +46,13 @@
     <artifact name="${org.apache.maven:maven-model}"/>
     <artifact name="${org.apache.maven:maven-model-builder}"/>
     <artifact name="${org.apache.maven:maven-repository-metadata}"/>
+    <artifact name="${org.apache.maven:maven-resolver-provider}"/>
+    <artifact name="${org.apache.maven:maven-settings}"/>
+    <artifact name="${org.apache.maven:maven-settings-builder}"/>
     <artifact name="${org.apache.maven.resolver:maven-resolver-api}"/>
     <artifact name="${org.apache.maven.resolver:maven-resolver-connector-basic}"/>
     <artifact name="${org.apache.maven.resolver:maven-resolver-impl}"/>
     <artifact name="${org.apache.maven.resolver:maven-resolver-named-locks}"/>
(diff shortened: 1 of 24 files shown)


TomasHofman:
I thought I created this PR already, but probably got interrupted and didn't push the button.

TomasHofman:
It will not compile because it depends on https://github.com/wildfly/wildfly-channel/pull/402 (SNAPSHOT version of wildfly-channel, compiling locally now).

TomasHofman:
@pedro-hos making this ready for review, introducing:

* The `--maven-settings` parameter to point prospero to the maven settings.xml file (defaults to ~/.m2/settings.xml). The settings file is then used by prospero to extract authentication credentials for Maven repositories.
* The `--filter-manifest`, that adds the ability to add an additional method of filtering available artifacts. This is used in case that the main input channel uses versionPatterns or no versions at all and resolves to the latest available artifact version. We would presently use it to filter out dependencies for specific EAP release (8.1 / 8.2) from a shared maven repo.
* Updates the wildfly-channel lib and prospero-metadata to versions that support accessing authenticated files over HTTP. This is used in cases when some channel metadata (manifest, blocklist) are referred as URLs rather, than Maven coordinates, and require HTTP authentication. The auth token can be provided as env variable ()`WILDFLY_CHANNEL_HTTP_AUTH_TOKEN` or system property (`org.wildfly.channel.http.auth.token`). See https://github.com/wildfly/wildfly-channel/pull/402

When we are OK re review, can we publish a new prospero release with this?

pedro-hos (approved):
Thank you very much @TomasHofman

PR #1039, opened 2026-05-11
Update GH-Action Workflow due version outdated #1038

### LINKED ISSUE #1038: Update GH-Action Workflow due version outdated
> ubuntu-latest-pmd
Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v2, actions/setup-java@v4. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

pedro-hos:
Closes https://github.com/wildfly/prospero/issues/1038

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,19 +6,25 @@ on:
       - main
     paths-ignore:
       - '.gitignore'
-      - 'examples'
+      - 'examples/**'
       - 'README.md'
   pull_request:
     branches-ignore:
       - 'future'
     paths-ignore:
       - '.gitignore'
-      - 'examples'
+      - 'examples/**'
       - 'README.md'
+
+permissions:
+  contents: read
+
+env:
+  MAVEN_ARGS: --show-version --batch-mode --no-transfer-progress
 
 jobs:
   build:
-    name: ${{ matrix.os }}-${{ matrix.jdk }}
+    name: Build (${{ matrix.os }} - JDK ${{ matrix.jdk }})
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -6,67 +6,86 @@ on:
...
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest ]
         jdk: [ 17, 21, 25 ]
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
mskacelik: checkout is already on v6

.github/workflows/main.yml
```diff
@@ -6,67 +6,86 @@ on:
...
         with:
           fetch-depth: 0
       - name: Set up JDK ${{ matrix.jdk }}
         uses: actions/setup-java@v4
```
mskacelik: setup-java is on v5

.github/workflows/main.yml
```diff
@@ -6,67 +6,86 @@ on:
...
         run: mvn clean install -DallTests -Pdist '-Dskip.container.tests=true'
-      - name: Build and test
+
+      - name: Build and test (Unix)
         if: runner.os != 'Windows'
         run: mvn clean install -DallTests -Pdist '-Dskip.container.tests=false'
-      - uses: actions/upload-artifact@v4
+
+      - name: Upload Surefire Reports
+        uses: actions/upload-artifact@v4
```
mskacelik: upload-artifact is on v7

.github/workflows/main.yml
```diff
@@ -6,67 +6,86 @@ on:
...
           path: '**/surefire-reports/*.txt'
+          retention-days: 7
+
   pmd:
     name: ubuntu-latest-pmd
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
mskacelik: ditto

.github/workflows/main.yml
```diff
@@ -6,67 +6,86 @@ on:
...
+          distribution: 'temurin'
           cache: maven
-      - name: Build and test
+
+      - name: Run PMD Check
         run: mvn pmd:aggregate-pmd-check -DskipTests
-      - uses: actions/upload-artifact@v4
+
+      - name: Upload PMD Report
+        uses: actions/upload-artifact@v4
```
mskacelik: ditto

.github/workflows/main.yml
```diff
@@ -6,67 +6,86 @@ on:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout repository
+        uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - name: Set up JDK
+      - name: Set up JDK 17
         uses: actions/setup-java@v4
```
mskacelik: ditto

mskacelik (commented):
Last note, Its usually a good practice to add a `MVN` environmental variable, something like:
```
env:
      MVN: ./mvnw --show-version --batch-mode --no-transfer-progress
```
And in the actual steps:
```yml
run: $MVN clean install # some additional flags
```

Note the use of **maven wrapper**, that is also I believe a common good practice, but AFAICT, prospero does not have a maven wrapper yet. Tho I am not 100% sure if you can use the non `.bat` wrapper for windows jobs.

.github/workflows/main.yml
TomasHofman: +1, let's use the latest versions available...

pedro-hos:
Thank you @mskacelik and @TomasHofman Can you please review again?

.github/workflows/main.yml
```diff
@@ -6,67 +6,89 @@ on:
...
       - '.gitignore'
-      - 'examples'
+      - 'examples/**'
       - 'README.md'
 
+permissions:
+  contents: read
+
+env:
+  MVN: mvn --show-version --batch-mode --no-transfer-progress
```
mskacelik: Ok, now I am retracting myself from the previous comment :), since we use Windows job, we cannot really use `$MVN`, we have to use the `${{ env.MVN }}` (like you just did), while this is fine and totally ok, maybe we can in theory only set `MAVEN_ARGS` which is by default included by maven (since 3.9.0) https://maven.apache.org/configure.html#maven_args-environment-variable.

And both `ubuntu-latest`, and `windows-latest` should be using `3.9.z` version of maven (we can actually check it with the current jobs thanks to `--show-version` flag :-))

So setting the env as:
```yaml
env:
  MAVEN_ARGS: --show-version --batch-mode --no-transfer-progress
```
Might be enough, hopefully (and than just using the normal `mvn clean install ...` should be in theory enough).
Worst case scenario, we could always do:
```sh
mvn ${{ env.MAVEN_ARGS }} clean instal ...
```

.github/workflows/main.yml
```diff
@@ -6,67 +6,89 @@ on:
...
       - name: Display report
         if: failure()
-        run: cat target/pmd.xml
-
+        run: |
+          if [ -f target/pmd.xml ]; then
+            cat target/pmd.xml
+          else
+            echo "PMD report not found"
+          fi
```
mskacelik, line 94: I believe its always better to use bash's `[[` instead of normal posix one `[` :-) (if you are in bash env) - note that you can also make it as one liner.
```suggestion
        run: [[ -f target/pmd.xml ]] && cat target/pmd.xml || echo 'PMD report not found'

```

mskacelik (approved):

TomasHofman (approved):
👍

PR #1042, opened 2026-05-14
[WFCORE-6934] Server startup marker file is not a reliable method to detect when the server is running

elguardian:
depends on https://github.com/wildfly/wildfly-core/pull/6756/
Issue: https://github.com/wildfly/wildfly-core/pull/6756/


TomasHofman:
So instead of making a decision based on the existence of a marker file (which can remain existing after the server process is killed), we will now make a decision based on whether the marker file is being locked. I understand the lock would be released by OS when the owner process is killed.

Looks like an improvement to me.

We are still waiting for how the wildfly-core PR review turns out, as we depend on that.

PR #1046, opened 2026-05-21
Allow prospero feature-pack to be provisioned with EE-10 and EE-11

jfdenise:


dist/wildfly-galleon-pack/src/main/config/wildfly-user-feature-pack-build.xml
```diff
@@ -22,9 +22,9 @@
   ~ 02110-1301 USA, or see the FSF site: http://www.fsf.org.
   -->
 
-<build xmlns="urn:wildfly:feature-pack-build:3.2" producer="org.wildfly.prospero:prospero-wildfly-galleon-pack">
+<build xmlns="urn:wildfly:feature-pack-build:3.5" producer="org.wildfly.prospero:prospero-wildfly-galleon-pack">
   <dependencies>
-    <dependency group-id="${prospero.base.feature-pack.groupId}" artifact-id="${prospero.base.feature-pack.artifactId}">
+    <dependency group-id="${prospero.base.feature-pack.groupId}" artifact-id="${prospero.base.feature-pack.artifactId}" allowedFamily="wildfly:jakarta-ee-10+">
```
pedro-hos, line 27: Do you think that's valid provide the `allowedFamily="wildfly:jakarta-ee-10+"` as environment parameter as we have done in  `artifact-id=` and `group-id=`?

pedro-hos (commented):
LGTM. The only point was the environment parameter . But I think that we can merge anyway.

dist/wildfly-galleon-pack/src/main/config/wildfly-user-feature-pack-build.xml
TomasHofman: We can experiment with that, it probably should work to have that as maven property if it works for the other attributes.

I would do that as a follow-up PR though.

TomasHofman (approved):
Thanks @jfdenise !

PR #1049, opened 2026-07-17
JBEAP-33965 - Internal urls exposed in prospero

pedro-hos:
Issue: https://redhat.atlassian.net/browse/JBEAP-33974


TomasHofman (commented):
Discussed over slack - I believe this would not resolve the problem, the internal URL must not be part of the sources at all. Should be provided in the infra via custom Maven settings.xml or system property or something like that.

TomasHofman:
Another way could be to ask prod to remove the URL from sources during productization. For instance EAP contains internal maven repo URLs as well in pom.xml, and it is removed during prod I believe.

pedro-hos:
We'll just remove the URL from sources during productization. For instance EAP contains internal maven repo URLs as well in pom.xml, and it is removed during productization.

PR #1050, opened 2026-07-20
Update to 1.4.0

pedro-hos:


pom.xml
```diff
@@ -63,7 +63,7 @@
         <connection>scm:git:git@github.com:wildfly/prospero.git</connection>
         <developerConnection>scm:git:git@github.com:wildfly/prospero.git</developerConnection>
         <url>https://github.com/wildfly/prospero</url>
-        <tag>HEAD</tag>
+        <tag>1.4.0.Final</tag>
```
TomasHofman: I'm not 100% sure, but 1.4.0.Final looks like improbable value here (for 1.4.1.Final-SNAPSHOT version). Should perhaps stay on HEAD? Do you think the plugin didn't change it back?

PR #1052, opened 2026-08-05
[JBEAP-34144] - prospero should not print error if user reject the update

pedro-hos:
Issue: https://redhat.atlassian.net/browse/JBEAP-34144

### DIFF
--- a/integration-tests/src/test/java/org/wildfly/prospero/it/cli/UpdateToVersionTest.java
+++ b/integration-tests/src/test/java/org/wildfly/prospero/it/cli/UpdateToVersionTest.java
@@ -218,7 +218,7 @@ public class UpdateToVersionTest extends CliTestBase {
                 ),
                 (ExecutionUtils.ExecutionResult e) -> {
                     try {
-                        e.assertReturnCode(ReturnCodes.PROCESSING_ERROR);
+                        e.assertReturnCode(ReturnCodes.SUCCESS);
                         assertThat(e.getCommandOutput())
                                 .contains("The update will downgrade following channels:")
                                 .contains("  * test-channel: 1.0.1 (Logical version 1.0.1)  ->  1.0.0 (Logical version 1.0.0)");
--- a/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
+++ b/prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
@@ -155,30 +155,34 @@ public class UpdateCommand extends AbstractParentCommand {
             Path targetDir = null;
             try {
                 targetDir = Files.createTempDirectory("update-candidate");
-                if (buildUpdate(updateAction, targetDir, console::confirmUpdates)) {
-                    console.println("");
-                    console.buildUpdatesComplete();
-
-                    ApplyCandidateAction applyCandidateAction = actionFactory.applyUpdate(installDir, targetDir);
-                    final List<FileConflict> conflicts = applyCandidateAction.getConflicts();
-                    if (!conflicts.isEmpty()) {
-                        FileConflictPrinter.print(conflicts, console);
-
-                        if (noConflictsOnly) {
-                            throw CliMessages.MESSAGES.cancelledByConfilcts();
-                        }
-
-                        if (!yes && !console.confirm(CliMessages.MESSAGES.continueWithUpdate(), "",
-                                CliMessages.MESSAGES.updateCancelled())) {
-                            return false;
-                        }
-                    }
-
-                    console.println(CliMessages.MESSAGES.applyingUpdates());
-                    applyCandidateAction.applyUpdate(ApplyCandidateAction.Type.UPDATE);
-                } else {
+                BuildResult result = buildUpdate(updateAction, targetDir, console::confirmUpdates);
+                if (result == BuildResult.ERROR) {
                     return false;
                 }
+                if (result != BuildResult.BUILT) {
+                    return true;
+                }
+
+                console.println("");
+                console.buildUpdatesComplete();
+
+                ApplyCandidateAction applyCandidateAction = actionFactory.applyUpdate(installDir, targetDir);
+                final List<FileConflict> conflicts = applyCandidateAction.getConflicts();
+                if (!conflicts.isEmpty()) {
+                    FileConflictPrinter.print(conflicts, console);
+
+                    if (noConflictsOnly) {
+                        throw CliMessages.MESSAGES.cancelledByConfilcts();
+                    }
+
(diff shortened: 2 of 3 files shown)


integration-tests/src/test/java/org/wildfly/prospero/it/cli/UpdateToVersionTest.java
```diff
@@ -218,7 +218,7 @@ public void downgradeMultiChannelRejected() throws Exception {
                 ),
                 (ExecutionUtils.ExecutionResult e) -> {
                     try {
-                        e.assertReturnCode(ReturnCodes.PROCESSING_ERROR);
+                        e.assertReturnCode(ReturnCodes.SUCCESS);
```
TomasHofman, line 221: In this use case, is the user asked for confirmation or is the operation terminated without asking?

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
```diff
@@ -155,30 +155,34 @@ private boolean performUpdate(UpdateAction updateAction, CliConsole console, Pat
...
-                                CliMessages.MESSAGES.updateCancelled())) {
-                            return false;
-                        }
+                BuildResult result = buildUpdate(updateAction, targetDir, console::confirmUpdates);
+                if (result == BuildResult.ERROR) {
+                    return false;
+                }
+                if (result != BuildResult.BUILT) {
+                    return true;
+                }
```
TomasHofman, line 164: In this case it would appear the `result` is either CANCELLED or NO_UPDATES, and we return true. (Probably OK.)

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
```diff
@@ -155,30 +155,34 @@ private boolean performUpdate(UpdateAction updateAction, CliConsole console, Pat
...
+                        throw CliMessages.MESSAGES.cancelledByConfilcts();
                     }
 
-                    console.println(CliMessages.MESSAGES.applyingUpdates());
-                    applyCandidateAction.applyUpdate(ApplyCandidateAction.Type.UPDATE);
-                } else {
-                    return false;
+                    if (!yes && !console.confirm(CliMessages.MESSAGES.continueWithUpdate(), "",
+                            CliMessages.MESSAGES.updateCancelled())) {
+                        return false;
```
TomasHofman, line 180: In this case we return false even though it was user choice to cancel, because we want to treat this "cancelled due to conflicts" as error, unlike the case above where on CANCELLED we return true?

TomasHofman: (Can be fine, just making sure if I got the reasoning...)

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/UpdateCommandTest.java
```diff
@@ -655,6 +655,82 @@ public void rejectDowngradeWithoutExplicitCommand() throws Exception {
...
+                CliConstants.DIR, installationDir.toAbsolutePath().toString(),
+                CliConstants.YES);
+
+        assertEquals(ReturnCodes.SUCCESS, exitCode);
+        assertEquals(0, getAskedConfirmation());
+        Mockito.verify(updateAction, never()).buildUpdate(any());
+        assertThat(getStandardOutput())
+                .contains("The update would introduce one or more unexpected channel downgrades.")
+                .contains("test-channel: 1.0.1  ->  1.0.0");
+    }
```
TomasHofman, line 732: I have question on this case - was there any user input here, or the tool shuts down by itself? How would the user enable the downgrade here?

TomasHofman: I'm thinking if there was no user input and the tool shut down because it detected a downgrade, maybe we should return ERROR here?

TomasHofman (commented):
I think nicely done, just have few questions...

integration-tests/src/test/java/org/wildfly/prospero/it/cli/UpdateToVersionTest.java
pedro-hos: The user is asked for confirmation. The test uses --manifest-versions=test-channel::1.0.0, so the downgrade is explicitly requested (not unexpected). The flow goes through hasUnexpectedDowngrade() → false, then prompts the user for confirmation, and the user responds "n". Since this is a user-initiated cancellation, it returns SUCCESS.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
pedro-hos: Correct. Both cases are not errors, either the user chose to cancel or there's nothing to update. Returning true (SUCCESS) is the intended behavior.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
pedro-hos: Good catch on the inconsistency. This return false is pre-existing behavior that was not changed in this PR. I agree it could be aligned in a follow-up change. Do you think that we'd change this behaviour as well?.

prospero-cli/src/test/java/org/wildfly/prospero/cli/commands/UpdateCommandTest.java
pedro-hos: You're right. There's no user input, the tool detects an unexpected downgrade and stops on its own. `Prepare.call()` was not handling `BuildResult.ERROR` and always fell through to `return ReturnCodes.SUCCESS`. Fixed it: `Prepare.call()` now returns `PROCESSING_ERROR` when `buildUpdate()` returns `ERROR`, and the test was renamed to `unexpectedDowngradeInPrepareReturnsError` with the assertion updated accordingly. I'll finish the tests and make the push here.

integration-tests/src/test/java/org/wildfly/prospero/it/cli/UpdateToVersionTest.java
TomasHofman: I guess that looks OK.

prospero-cli/src/main/java/org/wildfly/prospero/cli/commands/UpdateCommand.java
TomasHofman: OK, let's not change that here if that was existing behavior.

PR #1055, opened 2026-09-01
Update FeaturesAddAction to enable authenticated maven repositories

TomasHofman:
Not all functionalities are able to use authenticated maven repositories. This is use case uncovered by EAP installer MyFacesJsfLibrarySetupTest.

See https://github.com/jbossas/eap-installer/pull/394/

### DIFF
--- a/prospero-common/src/main/java/org/wildfly/prospero/actions/FeaturesAddAction.java
+++ b/prospero-common/src/main/java/org/wildfly/prospero/actions/FeaturesAddAction.java
@@ -87,6 +87,7 @@ public class FeaturesAddAction {
     private final Console console;
     private final CandidateActionsFactory candidateActionsFactory;
     private final FeaturePackTemplateManager featurePackTemplateManager;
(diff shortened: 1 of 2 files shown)


prospero-common/src/main/java/org/wildfly/prospero/actions/FeaturesAddAction.java
```diff
@@ -615,42 +619,45 @@ private Map<String, Set<String>> getAllLayers(FeaturePackLocation fpl)
...
+                                layersMap.put(model, names);
+                            }
+                            names.add(layer.getName());
                         }
-                        names.add(layer.getName());
                     }
                 }
             }
-        }
+            return layersMap;
```
pedro-hos, line 648: There's an extra blank line between return layersMap; and the closing brace.

pedro-hos (approved):
LGTM.  Only the extra blank line between return layersMap; and the closing brace. But, we can merge as it's. I don't think that will broke the check-style.

PR #1057, opened 2026-09-04
Enable setting cyclonedx properties in galleon

TomasHofman:
These are properties that control whether SBOM should be generated during provisioning, and whether prospero should fail on SBOM generation failure.


jfdenise (approved):

pedro-hos (approved):
LGTM, thank you Tomas. I'll merge it now

